PR #35, opened 2018-12-21
Improve installation scripts

mauriciovasquezbernal:
This PR contains some small fixes to the installation scripts and the docker build process


Documentation/components/k8s/developers.rst
```diff
@@ -7,37 +7,7 @@ Creating the Docker Images
 Docker 18.06 is needed to build the images, and the daemon has to be started with the `--experimental` flag.
 `See this issue to have more information <https://github.com/moby/moby/issues/32507>`_.
 
-You also need to ``export DOCKER_BUILDKIT=1`` before building the image
-
-::
-
-    docker build --build-arg KUBERNETES=true -t polycubenetwork/k8s-pod-network -f Dockerfile .
-    docker push polycubenetwork/k8s-pod-network
```
acloudiator: @mauriciovasquezbernal my recommendation is to provide CLI in a generic way. 
E.g.:
```
docker build --build-arg KUBERNETES=true -t docker_repo/name:tag -f Dockerfile .
docker push docker_repo/name:tag
```

frisso: Not clear to me:
1) why this has to be deleted
2) the meaning of the "export DOCKER_BUILDKIT...". When there are commands such that, I would suggest to add a brief sentence telling people what those are about.

mauriciovasquezbernal: I agree

mauriciovasquezbernal: No sure what you mean by deleted.
That comment is already present, please look at the full commit.

scripts/install.sh
```diff
@@ -27,6 +27,25 @@ function success_message {
 }
 trap error_message ERR
 
+function show_help() {
+usage="$(basename "$0") [mode]
+Polycube installation script
+
+mode:
+  pcn-iptables: install pcn-iptables service and related components
```
mbertrone: Maybe `only install pcn-iptables` and `only install pcn-k8s` instead of current message.

mauriciovasquezbernal:
@acloudiator @frisso could you please have a look?

PR #39, opened 2019-01-08
pcn-iptables: fix HORUS when all rules are removed

mbertrone:
when all rules are removed, and HOURS is enabled, parser reloading is needed.

### DIFF
--- a/src/services/pcn-iptables/src/Chain.cpp
+++ b/src/services/pcn-iptables/src/Chain.cpp
@@ -443,6 +443,8 @@ void Chain::updateChain() {
    * miss -> goto all pipeline steps
    */
 
+  parent_.ddos_mitigator_runtime_enabled_ = false;
+
   // Apply DDoS Mitigator optimization only if we are updating INPUT chain
   if ((name == ChainNameEnum::INPUT) && (parent_.horus_enabled)) {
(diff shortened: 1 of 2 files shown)


src/services/pcn-iptables/src/Chain.cpp
```diff
@@ -509,6 +511,27 @@ void Chain::updateChain() {
       }
     }
   }
+  if ( parent_.ddos_mitigator_runtime_enabled_ == false ) {
```
mauriciovasquezbernal: Please check the style here.

mbertrone: Applied clang-format. Thanks!

src/services/pcn-iptables/test/local_test_horus1.sh
```diff
@@ -0,0 +1,41 @@
...
+
+polycubectl pcn-iptables set horus=ON
+
+ping $ip -W 1 -c 2 -W 2
+
+pcn-iptables -A INPUT -s $ip -j DROP
+
+test_fail ping $ip -W 1 -c 2 -W 2
+
+pcn-iptables -D INPUT -s $ip -j DROP
```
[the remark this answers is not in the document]
mbertrone: Agree. Added in 4b0c73bfd56c49f2fb74398cc5db57b52c1716b9

PR #40, opened 2019-01-08
pcn-iptables: cleanup old references to DDoS in HORUS

mbertrone:
this commit removes old references to ddos mitigator, that was the original name of the module used to implement HORUS optimization.


mbertrone:
Thanks for the suggestion @mauriciovasquezbernal !
Done with last commit f5ba8ae

Issue #7, opened 2018-12-17
pcn-iptables: no counters for the default action when there are no rules in the chain


acloudiator:
pcn-iptables adopts some optimizations to skip the pipeline when no rules are instantiated in the target chain.
In particular, the Chain Logic module sets the PASS_LABELING label, which is used by the subsequent modules to understand the situation and act consequently.
However, I noticed that in this case, even though the system behaves correctly, it doesn't update the counters for the default action that is matched.

The issue can be easily reproduced by starting pcn-iptables without any rule and then send traffic to the INPUT or FORWARD chain.

- Sebastiano

mbertrone:
If no rules in `INPUT` and `FORWARD` chain, we apply optimization to skip `ChainLogic`, applying `PASS_LABELING` label to the packet.

IMO we have 3 possible solutions:
A) Keep this optimization, without update counters.
B) Keep this optimization, updating `INPUT` counters, that could be not always correct. E.g. half traffic goes through `INPUT` and half goes to `FORWARD`, so we have wrong counters.
C) Remove this optimization, with some performance loss, and update counters independently for `INPUT` and `FORWARD` chains, after [1] ChainLogic lookup.

@goldenrye [2], suggested to keep track of the issue, without removing optimization (A).

Any further suggestions?

[1] https://github.com/polycube-network/polycube/blame/106ece0232987bec295bda6419f1151aaf3473a8/src/services/pcn-iptables/src/datapaths/Iptables_ChainSelector_dp.c#L154
[2] https://github.com/polycube-network/polycube/pull/41#discussion_r246864901

PR #41, opened 2019-01-08
fix no counter when PASS_LABELING set (issue #7)

goldenrye:


src/services/pcn-iptables/src/datapaths/Iptables_ChainSelector_dp.c
```diff
@@ -132,6 +132,8 @@ static int handle_rx(struct CTXTYPE *ctx, struct pkt_metadata *md) {
   #if _INGRESS_ALLOWLOGIC
       pcn_log(ctx, LOG_DEBUG, "INGRESS LOGIC PASS. No rules for INPUT and FORWARD, and default is ACCEPT. ");
       updateForwardingDecision(PASS_LABELING);
+      incrementDefaultCountersInput(md->packet_len);
```
mbertrone: Update `INPUT` counters is not always correct since some pkts could belong to `FORWARD` chain.

src/services/pcn-iptables/src/datapaths/Iptables_ChainSelector_dp.c
```diff
@@ -132,6 +132,8 @@ static int handle_rx(struct CTXTYPE *ctx, struct pkt_metadata *md) {
...
+      _DEFAULTACTION_INPUT
```
mbertrone: Not useful, if `_INGRESS_ALLOWLOGIC` is `true`, `_DEFAULT_ACTION_INPUT` will be replaced with [1]

[1] https://github.com/polycube-network/polycube/blob/106ece0232987bec295bda6419f1151aaf3473a8/src/services/pcn-iptables/src/modules/ChainSelector.cpp#L62

src/services/pcn-iptables/src/datapaths/Iptables_ChainSelector_dp.c
goldenrye: Then I think just mention it in release note withtout fixing it, in this scenario counting and optimization seems to exclude each other. or we can add a option to allow user to pick up their choice.

mbertrone:
I think in case `no rules for INPUT/FORWARD` chains, we could not maintain counters and mention it in release notes.
The problem I can see, with introducing a new API to configure this behavior, is that we break `iptables` standard API and we need to document it.

mbertrone:
Issue tracked in #7 
If no further comments, I would close this PR.

PR #43, opened 2019-01-10
polycubed: remove methods declaration in rest server

mauriciovasquezbernal:
Just a small cleanup.


acloudiator:
Hi Folks, What was this cleanup for?

mauriciovasquezbernal:
I cannot remember the exact scope of those two declarations, this is quite old code and we can forget about it.

PR #45, opened 2019-01-11
pcn-iptables: implement support for interface filtering (-i -o options)

mbertrone:
Please don't merge until https://github.com/polycube-network/iptables/pull/1 is merged.
After that, we can update submodule reference in this PR.

this commit introduces support for filtering packets on interfaces.
e.g.
pcn-iptables -A INPUT -i eth0 -d 10.0.0.1 -j DROP
pcn-iptables -A OUTPUT -o eth1 -j DROP
pcn-iptables -A FORWARD -i veth1 -j ACCEPT

limitations:
- on FORWARD chain only support for -i parameter.
Due to architecture of pcn-iptables, support -o parameter would require
to guess linux routing decision in INGRESS hook, or to extend EGRESS hook
to filter also packets already processed by FORWARD pipeline.

### DIFF
--- a/src/services/pcn-iptables/src/Iptables.cpp
+++ b/src/services/pcn-iptables/src/Iptables.cpp
@@ -487,3 +487,13 @@ bool Iptables::fibLookupEnabled() {
 
   return fib_lookup_enabled_;
 }
+
+uint16_t Iptables::interfaceNameToIndex(const std::string &interface_string) {
+  try {
+    auto p = get_port(interface_string);
+    int index = p->index();
+    return index;
+  } catch (std::exception) {
+    throw std::runtime_error("invalid interface name");
+  }
+}
--- a/src/services/pcn-iptables/src/Iptables.cpp
+++ b/src/services/pcn-iptables/src/Iptables.cpp
@@ -489,11 +489,7 @@ bool Iptables::fibLookupEnabled() {
 }
 
 uint16_t Iptables::interfaceNameToIndex(const std::string &interface_string) {
-  try {
-    auto p = get_port(interface_string);
-    int index = p->index();
-    return index;
-  } catch (std::exception) {
-    throw std::runtime_error("invalid interface name");
-  }
+  auto p = get_port(interface_string);
+  int index = p->index();
+  return index;
 }
--- a/src/services/pcn-iptables/src/Utils.cpp
+++ b/src/services/pcn-iptables/src/Utils.cpp
@@ -432,7 +432,7 @@ bool Chain::transportProtoFromRulesToMap(
 bool Chain::portFromRulesToMap(
         const uint8_t &type, std::map<uint16_t, std::vector<uint64_t>> &ports,
         const std::vector<std::shared_ptr<ChainRule>> &rules) {
-  std::vector<uint16_t> dont_care_rules;
+  std::vector<uint32_t> dont_care_rules;
 
   uint32_t rule_id;
   uint16_t port;
@@ -483,7 +483,7 @@ bool Chain::interfaceFromRulesToMap(
         const uint8_t &type, std::map<uint16_t, std::vector<uint64_t>> &interfaces,
         const std::vector<std::shared_ptr<ChainRule>> &rules,
         Iptables& iptables) {
-  std::vector<uint16_t> dont_care_rules;
+  std::vector<uint32_t> dont_care_rules;
 
   uint32_t rule_id;
   uint16_t interface;
--- a/src/services/pcn-iptables/src/Utils.cpp
+++ b/src/services/pcn-iptables/src/Utils.cpp
@@ -479,6 +479,61 @@ bool Chain::portFromRulesToMap(
   return brk;
 }
 
+bool Chain::interfaceFromRulesToMap(
+        const uint8_t &type, std::map<uint16_t, std::vector<uint64_t>> &interfaces,
+        const std::vector<std::shared_ptr<ChainRule>> &rules,
+        Iptables& iptables) {
+  std::vector<uint16_t> dont_care_rules;
+
+  uint32_t rule_id;
+  uint16_t interface;
+
+  bool brk = true;
+
+  for (auto const &rule : rules) {
+    try {
+      rule_id = rule->getId();
+      interface = 0;
+      if (type == IN_TYPE) {
+        std::string interface_string = rule->getInIface();
+        interface = iptables.interfaceNameToIndex(interface_string);
+      } else {
+        std::string interface_string = rule->getOutIface();
+        interface = iptables.interfaceNameToIndex(interface_string);
+      }
+    } catch (std::runtime_error re) {
+      // Interface not set: don't care rule.
+      dont_care_rules.push_back(rule_id);
+      continue;
+    }
+
+    auto it = interfaces.find(interface);
+    if (it == interfaces.end()) {
+      // First entry
+      std::vector<uint64_t> bitVector(
+              FROM_NRULES_TO_NELEMENTS(Iptables::max_rules_));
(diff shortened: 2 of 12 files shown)


src/services/pcn-iptables/src/Iptables.cpp
```diff
@@ -486,4 +486,14 @@ bool Iptables::fibLookupEnabled() {
...
\ No newline at end of file
+}
+
+uint16_t Iptables::interfaceNameToIndex(const std::string &interface_string) {
+  try {
+    auto p = get_port(interface_string);
+    int index = p->index();
+    return index;
+  } catch (std::exception) {
+    throw std::runtime_error("invalid interface name");
```
mauriciovasquezbernal: Is this necessary to catch and rethrow the exception?
The problem I see is that we are hiding the real exception, why not just to remove the whole catch block and let the exception go?

src/services/pcn-iptables/src/Utils.cpp
```diff
@@ -479,6 +479,61 @@ bool Chain::portFromRulesToMap(
...
+      rule_id = rule->getId();
+      interface = 0;
+      if (type == IN_TYPE) {
+        std::string interface_string = rule->getInIface();
+        interface = iptables.interfaceNameToIndex(interface_string);
+      } else {
+        std::string interface_string = rule->getOutIface();
+        interface = iptables.interfaceNameToIndex(interface_string);
+      }
+    } catch (std::runtime_error re) {
```
mauriciovasquezbernal: Why just catching `std::runtime_error`, I think in this case using `...` could be better.

src/services/pcn-iptables/src/defines.h
```diff
@@ -54,9 +54,10 @@ const uint8_t IPDESTINATION = 2;
...
 const uint8_t PORTSOURCE = 4;
 const uint8_t PORTDESTINATION = 5;
-const uint8_t TCPFLAGS = 6;
-const uint8_t BITSCAN = 7;
-const uint8_t ACTION = 8;
+const uint8_t INTERFACE = 6;
+const uint8_t TCPFLAGS = 7;
+const uint8_t BITSCAN = 8;
+const uint8_t ACTION = 9;
 }
```
mauriciovasquezbernal: I know this is out of scope of this PR. Wouldn't it make sense to use an enum in this case?

src/services/pcn-iptables/src/Iptables.cpp
mbertrone: The only reason is to provide a cleaner message to the user, in case of error, if a wrong interface name is specified.
If you write something like `pcn-iptables -A INPUT -i wrong_iface_name -j ACCEPT`, you will get:
A) `mat::at` (without throw)
B) `invalid interface name` (with throw)

src/services/pcn-iptables/src/Utils.cpp
mbertrone: In both `IN_TYPE` and `OUT_TYPE`, exception is raised by `getIn/OutInterface()` if this field is not set, and is raised by `interfaceNameToIndex()` if interface name is not valid.
If exception is raised in one of the previous case it means I have to consider current rule, as a don't care rule `*`.

This block can be replaced by a conditional, but this requires to change the pattern we use to handle this cases.
Do you think it is worth? What benefits can you see?

src/services/pcn-iptables/src/defines.h
mbertrone: You are right. `enum` would be a cleaner solution.
I'll keep track of that, and update in a following PR.

src/services/pcn-iptables/src/Iptables.cpp
mbertrone: After #53 merged, we can remove this try-catch block

src/services/pcn-iptables/src/Utils.cpp
mbertrone: After an offline talk with Mauricio, we agreed using exceptions for control flow handling is not the best practice.
Furthermore, this improvement will require many changes to the code, so it will be part of a  separate PR.

src/services/pcn-iptables/src/Utils.cpp
```diff
@@ -479,6 +479,61 @@ bool Chain::portFromRulesToMap(
   return brk;
 }
 
+bool Chain::interfaceFromRulesToMap(
+        const uint8_t &type, std::map<uint16_t, std::vector<uint64_t>> &interfaces,
+        const std::vector<std::shared_ptr<ChainRule>> &rules,
+        Iptables& iptables) {
+  std::vector<uint16_t> dont_care_rules;
+
+  uint32_t rule_id;
```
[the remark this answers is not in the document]
mbertrone: Fixed in c97c2f7

src/services/pcn-iptables/src/Utils.cpp
```diff
@@ -479,6 +479,61 @@ bool Chain::portFromRulesToMap(
...
+  if (interfaces.size() != 0 && dont_care_rules.size() != 0) {
+    std::vector<uint64_t> bitVector(
+            FROM_NRULES_TO_NELEMENTS(Iptables::max_rules_));
+    interfaces.insert(std::pair<uint16_t, std::vector<uint64_t>>(0, bitVector));
+    for (auto const &ruleNumber : dont_care_rules) {
+      for (auto &interface : interfaces) {
+        SET_BIT((interface.second)[ruleNumber / 63], ruleNumber % 63);
+      }
+    }
+    brk = false;
```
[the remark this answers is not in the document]
mbertrone: As explained here [1], programs injection prioritizes blocks that can possibility early break the pipeline, speeding up execution time.
If there is at least one wildcard rule, optimization is not possible [2].
If there are no interface rules, it doesn't make sense to compute the bitvector for wildcard rules, since the block of the pipeline will not be injected at all.

[1] https://github.com/polycube-network/polycube/blob/master/src/services/pcn-iptables/src/Chain.cpp#L542
[2] https://github.com/polycube-network/polycube/pull/45/files#diff-2e6cf64d7d08b4bc5410d248f0b38009R102

src/services/pcn-iptables/src/Iptables.cpp
mbertrone: Done in e24fc86.

src/services/pcn-iptables/src/modules/InterfaceLookup.cpp
```diff
@@ -0,0 +1,150 @@
...
+                        (chain == ChainNameEnum::OUTPUT)
+                            ? ProgramType::EGRESS
+                            : ProgramType::INGRESS) {
+  this->type_ = type;
+
+  auto it = ports.find(0);
+  if (it == ports.end()) {
+    wildcard_rule_ = false;
+    wildcard_string_ = "";
+    // std::cout << "-- wildcard NO --+" << std::endl;
```
frisso: I would delete this line (and the two debug messages below as well).

mbertrone: Why we should delete debug prints?
I think debug messages are always welcome. Btw I don't like `std::cout` approach, using logger is better, but in this method is not easy to get access to logger instance (maybe a point to consider for developer-friendly improvements)

frisso: Debug print are always welcome, but... debug means for the developer (this should be avoided in the final commit, as we should commit only code that has been properly tested), of for the user who may have an unexpected behavior?
In any case, I would suggest to use the logger.
So, for the time being I would remove this print, and create maybe a issue saying why logger may not be easy to use, so that we can track this issue.

PR #52, opened 2019-01-29
Add support for bpf_redirect_map in XDP Cubes

sebymiano:
This commit introduces the possibility to use the `bpf_redirect_map` helper for XDP programs. 
This helper has improved performance compared to the normal `bpf_redirect` function that is used for other programs.
It uses a new map called `BPF_DEVMAP` that contains a mapping between a virtual index (used internally by the `Cube` to indicate a specific port) and the `ifindex` of the physical interface connected to the `Cube` port.

This map can only be used for sending packets out on a netdev (either physical or virtual) but cannot be used for the Cube-to-Cube communication. 
The approach adopted here is to check the value in the `forward_chain_` map; if the value is 0 than there are no entries for that output port and the devmap is used.

This commit depends on the BCC PR https://github.com/polycube-network/bcc/pull/1, which adds a single line definition to the generic `BPF_TABLE`.

### DIFF
--- a/src/polycubed/src/cube_xdp.cpp
+++ b/src/polycubed/src/cube_xdp.cpp
@@ -33,7 +33,7 @@ CubeXDP::CubeXDP(const std::string &name,
     : Cube(name, service_name,
            PatchPanel::get_xdp_instance(),
            PatchPanel::get_tc_instance(),
-           level, type), attach_flags_(0) {
+           level, type, CubeXDP::MASTER_CODE), attach_flags_(0) {
   switch(type) {
     // FIXME: replace by definitions in if_link.h when update to new kernel.
     case CubeType::XDP_SKB:
@@ -44,9 +44,13 @@ CubeXDP::CubeXDP(const std::string &name,
       break;
   }
 
-  // it has to be done here becuase it needs the load, compile methods
+  // it has to be done here because it needs the load, compile methods
   // to be ready
   Cube::init(ingress_code, egress_code);
+  auto devmap_ = master_program_->get_devmap_table("xdp_devmap_");
+  xdp_devmap_ = std::unique_ptr<ebpf::BPFDevmapTable>(new ebpf::BPFDevmapTable(devmap_));
+
+  logger->debug("Created devmap with fd {0}", xdp_devmap_->get_fd());
 }
 
 CubeXDP::~CubeXDP() {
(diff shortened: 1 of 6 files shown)


src/polycubed/src/cube_xdp.cpp
```diff
@@ -44,9 +44,13 @@ CubeXDP::CubeXDP(const std::string &name,
...
   }
 
-  // it has to be done here becuase it needs the load, compile methods
+  // it has to be done here because it needs the load, compile methods
   // to be ready
   Cube::init(ingress_code, egress_code);
+  auto devmap_ = master_program_->get_devmap_table("xdp_devmap_");
+  xdp_devmap_ = std::unique_ptr<ebpf::BPFDevmapTable>(new ebpf::BPFDevmapTable(devmap_));
+
+  logger->debug("Created devmap with fd {0}", xdp_devmap_->get_fd());
```
mauriciovasquezbernal: Are this kind of debug messages really worthy?

src/polycubed/src/port.cpp
```diff
@@ -138,12 +138,24 @@ void Port::send_packet_out(const std::vector<uint8_t> &packet, Direction directi
...
+*/
 void Port::connect(Port &p1, Node &iface) {
   std::lock_guard<std::mutex> guard(p1.port_mutex_);
   p1.peer_iface_ = &iface;
   // TODO: provide an accessor for forward chain?
   // Even smarter, should be forward chain a property of Port?
-  p1.parent_.update_forwarding_table(p1.index_, iface.get_index());
+
+  if (p1.parent_.get_type() == CubeType::TC) {
+    p1.parent_.update_forwarding_table(p1.index_, iface.get_index(), true);
```
mauriciovasquezbernal: I think the last parameter should be `false`,  only `XDP` programs have `bpf_redirect_map` support.

src/polycubed/src/port.cpp
```diff
@@ -85,11 +87,9 @@ void Port::set_peer(const std::string &peer) {
     if (peer_ == peer) {
       return;
     }
+    peer_ = peer;
```
mauriciovasquezbernal: `peer_` is updated after `ServiceController::set_port_peer(*this, peer);` because it is only after that function finishes that the peer has actually been updated.

There are some problems of moving it there.
1. The logic to remove a previous connection is broken, it'll try to remove a new non existing connection while the old one will remain there. https://github.com/polycube-network/polycube/blob/master/src/polycubed/src/service_controller.cpp#L231
2. If `ServiceController::set_port_peer(*this, peer);` raises an exception the peer would be configured to a wrong value.

src/polycubed/src/cube_xdp.cpp
sebymiano: Agree, we can definitely remove this debug message.

src/polycubed/src/port.cpp
sebymiano: The last parameter tells if the connection is with a `netdev` or not.
To me, it is correct to pass the value `true` even for TC cubes but, in the current implementation they simply ignore that value and use the default implementation the `bpf_redirect_map` in the `Cube` class.
On the other hand, `XDP Cubes` extend that function to consider the last parameter writing the value in the `BPF_DEVMAP`.

What do you think?

frisso: It would be more clear to me if we create a define such as "IS_NETDEV true", so that the prototype of the function remains as is, but the readability of the code looks better: reader knows immediately what this parameter is about when it reads "IS_NETDEV" as param of this function.

frisso:
@sebymiano Is it now the right time to update this PR and merge in master?

sebymiano:
@frisso Yes, it would be nice to merge it into the master but unfortunately, after the changes on the `transparent services`, it requires a non-negligible refactory.
I'll synch with @mauriciovasquezbernal to understand better how to proceed.

acloudiator:
This code goes invalid, as it involves rework after transparent service code.

PR #54, opened 2019-01-30
Added in-daemon validation

Fylax:
The input data validation process has been moved from services to daemon.
This resulted in moving the whole REST server (pistache) into the daemon as well.

Since now the daemon requires run-time YANG parsing, a library named `libyang` has been added as dependency.

The validation process supports all the YANG features used by the current service, but it is not complete as it performs only static analysis (e.g., pattern validation, number in range check), while it lacks of dynamic analysis (e.g., min-elements). The only dynamic analysis currently supported is `when` statement validation (only a subset of `XPath` specification is supported, and it was not hard tested).

To better adhere to YANG specification, it has been added a new `extension` whithin the datamodel-common, named `init-only-config`. All nodes marked with this extension are configuration nodes that can be set only during initialization, so they cannot be updated later on.

Finally, all services with the exception of `bridge` (it wasn't ported to previous structure, so I was unable to update it) have been ported to newest architecture.


Fylax:
As per request, I tried to split a massive commit into smaller ones to make review easier.

Moreover, I will provide some documentation to give a general idea.
image

This component should help in a better understanding. In a nutshell, the main point of this work is the _Resource_. They are divided into _Body Resources_ and _Endpoint Resources_. The formers are in charge of request's body validation, while the latter validates the path parameters and handle HTTP requests.

For code reuse, _Endpoint Resources_ inherit from the _Body_ ones.

The distinction is due to the fact that resources inside a YANG `rpc` or `action` are body-only.

The whole validation process is made by a set of validators (registered to each resource).

Mainly (whether they are in body or endpoint namespace) there are two types of resources:
1. LeafResource: handles YANG `leaf`, performing validation on input scalar types
2. ParentResource: takes a complex JSON object as input and passes a subset of it to every one of his children.

Moreover, some specialized classes (such as ListResource) exists, and they are in charge to handle the correspondent YANG node (`list` in this case).

Finally, the whole tree is created by the Parser::Yang, that exploits an abstract factory to achieve decoupling with actual service communication channel (only SharedObject is supported with this PR, but adding gRPC from this work standpoint would just require to create the corresponding Factory and the associated _gRPCResources_.

CMakeLists.txt
```diff
@@ -5,6 +5,8 @@ project(polycube)
 set (CMAKE_CXX_STANDARD 11)
 
 include(cmake/GetGitRevisionDescription.cmake)
+list(APPEND CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/cmake")
+list(APPEND CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/cmake/nlohmann_json")
```
acloudiator: What is `nlohmann_json`?

cmake/nlohmann_json/Findnlohmann_json.cmake
```diff
@@ -0,0 +1,15 @@
+include(FindPackageHandleStandardArgs)
+set(${CMAKE_FIND_PACKAGE_NAME}_CONFIG ${CMAKE_CURRENT_LIST_FILE})
+find_package_handle_standard_args(nlohmann_json CONFIG_MODE)
```
acloudiator: `nlohmann_json`?! It seems it's coming from cmake output directory. My suggestion is to change it.

src/polycubed/src/polycubed_core.cpp
```diff
@@ -434,25 +423,33 @@ void PolycubedCore::detach(const std::string &cube_name,
...
+  Pistache::Http::Client client;
 
-  ServiceController &s = iter->second;
+  auto opts = Pistache::Http::Client::options();
+  client.init(opts);
+
+  auto port = std::to_string(config.getServerPort());
+  auto service_name = ServiceController::get_cube_service(cube);
+
+  std::string url("http://127.0.0.1:" + port + "/polycube/v1/" + service_name +
```
acloudiator: Will it always be `127.0.0.1`?
Any use case, where the IP will be external/remote and we'll have to use it? I believe we can set the IP and Port as an environment variable, is it? 
Please correct me if I'm wrong. Thanks!

CMakeLists.txt
mauriciovasquezbernal: It is the json library we use: https://github.com/nlohmann/json

cmake/nlohmann_json/Findnlohmann_json.cmake
mauriciovasquezbernal: What are the reasons to change it? What alternative do you suggest?

src/polycubed/src/polycubed_core.cpp
mauriciovasquezbernal: Yes, it is a local connection because it is starting from the same process, so 127.0.0.1 should work in all cases. Anyway this code has to be improved in the future because this is not good to use tcp to access to the same process.

mauriciovasquezbernal: This has already been improved, that static IP is not present in the code anymore.

cmake/nlohmann_json/Findnlohmann_json.cmake
frisso: @acloudiator Can you please be more specific? It would be nice to merge this big commit in order to move on. Thanks!

acloudiator: @frisso Following my conversation with @mauriciovasquezbernal this makes more sense and we've it under src/libs. I'm marking this conversation as resolved but still wait for @goldenrye feedback as this is a huge PR.

Fylax:
> @Fylax
> hard to understand this scaring commit, but one question is that this commit contains huge amount of code under the src/polycubed/src/server which is not specific to polycube, I don't know if it is necessary to compile into polycubed, or should we organize them under src/lib and compile into a dynamic library?

src/polycubed/src/server is where I put the whole in-daemon validation. I just kept it in a different folder to make things clearer. Its main purpose is to provide REST API based on data-model, incoming request validation and communication with the service.

Issue #1, opened 2018-12-17
pcn_log cannot be split in multiple lines


acloudiator:
Mauricio identified a weird limitation that can cause many troubles to developers.

pcn_log is not a function nor a macro, it is just a string that is looked and replaced in the code using a set of regular expressions (DatapathLog::parse_log)[1]. Currently, those regex are only able to replace pcn_log correctly if this is in a single line, otherwise strange compilation errors happen.

This limitation is because this is not possible to use a map inside a macro with bcc, unfortunately, I haven't been able to look a solution for it.

https://github.com/polycube-network/polycube/blob/master/src/polycubed/src/datapath_log.cpp#L211

PR #55, opened 2019-01-31
Allow pcn_log to be split in multiple lines

sebymiano:
This commit fixes a "bug" in the `pcn_log` function that did not allow to split the `pcn_log` body into multiple lines.
It uses an improved regex function to check the end of the function and replace it with the correct version so that compilation errors can be avoided.


frisso:
Not clear to me how it could be used. I suggest to add a brief comment in the documentation saying how to split the content over multiple lines, and also that this practice should be really avoided because it affects the capability to parse the output log with another program (which may expect one log per line).

mauriciovasquezbernal:
@frisso you're right, the text present on a logging primitive should be in a single line, your reason is a valid one and there is another one, it can be difficult to look for a error string in the code if this is broken into multiple lines.

However this is not the goal of this PR, the goal is to allow splitting the call of a  `pcn_log` in multiple lines, for example putting the text and the parameters in different ones. For example:

now:
```
pcn_log(ctx, LOG_TRACE, "Received new packet: in_port: %d eth_type: 0x%x mac_src: %M mac_dst: %M", md->in_port, bpf_ntohs(eth->proto), eth->src, eth->dst);
```

after this PR:
```
pcn_log(ctx, LOG_TRACE, 
        "Received new packet: in_port: %d eth_type: 0x%x mac_src: %M mac_dst: %M",
        md->in_port, bpf_ntohs(eth->proto), eth->src, eth->dst);
```

This is weird but the second case is not supported currently as described in https://github.com/polycube-network/polycube/issues/1.

frisso:
Got it, Mauricio, thank you very much.
I suggest to add this explanation in the commit description; otherwise I feel hard to understand what we're trying to do.

mauriciovasquezbernal:
I tested it and it works pretty good, there is just one case it doesn't cover (was also not covered on previous version)
```
pcn_log (....)
       ^
       Space
```

Would you mind applying the next diff to also supporting it?

Thanks so much for the PR.

```
diff --git a/Documentation/developers.rst b/Documentation/developers.rst
index 1938a89..c6f071c 100644
--- a/Documentation/developers.rst
+++ b/Documentation/developers.rst
@@ -155,7 +155,6 @@ Two special format specifiers are available:
 Please note the the custom specifiers spec the data to be in network byte order while standard specifiers expects it to be in host by order.
 
 Current limitations:
- - The whole call to `pcn_log` should be on the same line
  - Cannot be used inside a macro
  - Maximum 4 arguments are allowed
 
diff --git a/src/polycubed/src/datapath_log.cpp b/src/polycubed/src/datapath_log.cpp
index 98c04e1..a59b0e5 100644
--- a/src/polycubed/src/datapath_log.cpp
+++ b/src/polycubed/src/datapath_log.cpp
@@ -213,11 +213,11 @@ std::string DatapathLog::replace_string(std::string& subject, const std::string&
 static std::regex regComments(R"***((?:\/\/(?:\\\n|[^\n])*)|(?:\/\*[\s\S]*?\*\/)|((?:R"([^(\\\s]{0,16})\([^)]*\)\2")|(?:@"[^"]*?")|(?:"(?:\?\?'|\\\\|\\"|\\\n|[^"])*?")|(?:'(?:\\\\|\\'|\\\n|[^'])*?')))***");
 
 // matches all calls to pcn_log(ctx, ...)
-static std::regex regN(R"***(pcn_log\(([\s\S]+?)\)\s*;)***");
+static std::regex regN(R"***(pcn_log\s*\(([\s\S]+?)\)\s*;)***");
 static std::regex regNewLine(R"***(/(\r\n)+|\r+|\n+|\t+/i)***");
 static std::regex regAddSpaces(R"***( +)***");
 
-static std::regex regNPkt(R"***(pcn_pkt_log\(([\s\S]+?)\)\s*;)***");
+static std::regex regNPkt(R"***(pcn_pkt_log\s*\(([\s\S]+?)\)\s*;)***");
 
 std::string DatapathLog::dp_callback(const std::smatch& m) {
     std::string match = std::regex_replace(m.str(1), regNewLine, "");

```

sebymiano:
Thanks for the clarification @mauriciovasquezbernal.
I've applied your patch and slightly modified the commit message to make it more clear.

PR #57, opened 2019-02-06
This section can be removed

acloudiator:
It would be better to have separate CONTRIBUTORS file for this.

```
The following additional people are mentioned in commit logs as having provided
helpful bug reports, suggestions or have otherwise provided value to the
project:
```


frisso:
This doesn't look a very important issue; anyway, I personally would like to keep everything in a single file, although (as it is now) separating authors from contributors. Similar open-source projects share the same way.
Anyway, since this does not look important to me, I would not fight against this change if the others feel it's needed.

PR #61, opened 2019-02-08
Fix code style

mauriciovasquezbernal:
Create a tool to (scripts/check-style.py) check/fix the code style and then run it to fix the code style of the whole project. Datapath code style was also fixed as https://github.com/polycube-network/polycube/pull/55 prevented potential problems with `pcn_log`. Automatic code generated that is not intended to be touched by the user is ignored by the script.

This is a quite invasive change but I think this is better to do now, after it all PRs must have the code style right.


Documentation/developers.rst
```diff
@@ -348,7 +348,9 @@ Additional hints
 
 1. **Creating multiple data plane programs**. If possible, it would be nice to create a single dataplane program, and enabling/disabling some portions using conditional compilation macros.
 
-2. **Coding Style**: Please be sure that the coding style satisfies the rules defined in the [clang-format](.clang-format) file. In order to fix the style of a source file run `clang-format -style=file -i <file.[cpp,h]>` on the root folder of polycube.
+2. **Coding Style**: The `scripts/check-style.py` uses `clang-format` to check the code style.
+This tool has to be executed from the root folder, when not parameters are passed it checks all the code.
```
frisso: ==>
This tool has to be executed from the root folder.
A list of files or folders to check is received as argument; the entire source code is checked when no parameters are passed.
The `--fix` option will automatically fix the code style instead of simply checking it.

PR #62, opened 2019-02-08
pcn-router: remove arp table entries when deleting a port

francescomessina:
When a port is deleted, the ARP table must also be cleaned

### DIFF
--- a/src/services/pcn-router/src/Ports.cpp
+++ b/src/services/pcn-router/src/Ports.cpp
@@ -181,6 +181,17 @@ void Ports::removeEntry(Router &parent, const std::string &name) {
 
   auto router_port = parent.get_hash_table<uint16_t, r_port>("router_port");
 
+  // remove the ArpEntry from the datapath, for this port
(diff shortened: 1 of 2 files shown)


src/services/pcn-router/src/Ports.cpp
```diff
@@ -181,6 +181,18 @@ void Ports::removeEntry(Router &parent, const std::string &name) {
...
+  auto arp_entries = arp_table.get_all();
+  for (auto &entry : arp_entries) {
+    auto key = entry.first;
+    auto value = entry.second;
+    auto port = parent.get_port(value.port);
+
+    if (port->name() == name)
+      arp_table.remove(key);
+  }
+
```
mauriciovasquezbernal: It is not necessary to call `get_port` for each entry, we can compare if the entry is related to this port by using the port's index. Notice that `port` refers to the port that is going to be deleted.

```
if (port->index() == value.port) {
  arp_table.remove(key);
}
```

francescomessina: Ok, thanks @mauriciovasquezbernal.

PR #63, opened 2019-02-09
Bug and Feature Request template

acloudiator:
This is in process of making the repo more organized and documented and easy to navigate.
So please provide your feedback. Thanks!

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -24,8 +24,12 @@ A clear and concise description of what you expected to happen.
 Please share all the relevant logs, which can help the developers to diagnose the issue.
 
 **OS and Polycube Version Details (please complete the following information):**
-- OS: [e.g. Ubuntu 18.04 LTS]
-- Version [e.g. 0.1.0-alpha]
+- OS details: 
+1. OS: `cat /etc/*-release` 
+2. Kernel: `uname -r`
+- Polycube Details [e.g. 0.x.x]
+1. Version: `polycubectl version`
+2. Logs: `cat /var/log/polycube/polycubed.log`
 
 **Additional context**
 Add any other context about the problem here.
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -20,12 +20,12 @@ Steps to reproduce the behavior:
 **Expected behavior**
 A clear and concise description of what you expected to happen.
 
-**Screenshots**
-If applicable, add screenshots to help explain your problem.
+**Logs**
+Please share all the relevant logs, which can help the developers to diagnose the issue.
 
 **OS and Polycube Version Details (please complete the following information):**
 - OS: [e.g. Ubuntu 18.04 LTS]
- - Version [e.g. 0.1.0-alpha]
+- Version [e.g. 0.1.0-alpha]
 
 **Additional context**
 Add any other context about the problem here.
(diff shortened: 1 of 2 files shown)


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,31 @@
...
+1. Go to '...'
+2. Modify this '....'
+3. Execute this '....'
+4. See error
+
+**Expected behavior**
+A clear and concise description of what you expected to happen.
+
+**Screenshots**
+If applicable, add screenshots to help explain your problem.
```
mauriciovasquezbernal: Maybe screenshot are not the best idea, they are good for GUI applications, for us a dump of the logs is better (we can lookup text on them).

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,31 @@
...
+
+**OS and Polycube Version Details (please complete the following information):**
+- OS: [e.g. Ubuntu 18.04 LTS]
```
mauriciovasquezbernal: Small alignment problem here.

frisso: I would suggest the command to get OS and Polycube versions here, so that the user doesn't have to find them from himself.

acloudiator: @mauriciovasquezbernal I found only one in the last line for version, been sorted. 
@frisso That's thoughtful! Also, it would be better to include commands/path for logs to collect.

.github/ISSUE_TEMPLATE/bug_report.md
acloudiator: Changed to logs, but would be better to include the path/commands too.

.github/ISSUE_TEMPLATE/bug_report.md
frisso: @mauriciovasquezbernal : could you please add directions (1) to get logs, (2) polycube version and (3) os version? Thank you very much.

mauriciovasquezbernal: Sure, here we go:
kernel: `uname -r`
os: `cat /etc/*-release`
polycube: `polycubectl version`
logs: `cat /var/log/polycube/polycubed.log`

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,31 @@
+---
+name: Bug report
+about: Create a report to help us improve
+title: "[BUG]"
+labels: bug
+assignees: ''
+
+---
+
+**Describe the bug**
```
mauriciovasquezbernal: Have you checked the preview of this?
I think you are missing some line breaks to get a good formatting.

acloudiator: Can you exactly point where?! - that would be helpful.

mauriciovasquezbernal: I just noticed that you are not using headers at all. I think it could be more organized if you put things like "Describe the bug", "To reproduce" and so on in headers.

.github/ISSUE_TEMPLATE/bug_report.md
acloudiator: added

.github/ISSUE_TEMPLATE/bug_report.md
acloudiator: It's a generic template that we get from GitHub. For readability, I agree. 
This has been modified.

PR #64, opened 2019-02-11
Updates for check style script

mauriciovasquezbernal:
- add polycubed to default path
- return 0 when --fix is used
- remove old script


mauriciovasquezbernal:
>     1. Add author and copyright information in 'check-style.py'

I added a check for the copyright. I think the check for authors should be in another script.

>     2. Make sure that installation instructions include also Python (I would mark it as optional, telling also people that is required to run some scripts in case you're a developer).

Python is already there, actually not sure why.

src/services/pcn-firewall/src/Ports.h
```diff
@@ -1,14 +1,18 @@
...
-*
-* OpenAPI spec version: 2.0
-*
-* NOTE: This class is auto generated by the swagger code generator program.
-* https://github.com/polycube-network/swagger-codegen.git
-* branch polycube
-*/
-
+/*
+ * Copyright 2017 The Polycube Authors
```
frisso: 2017 --> 2018.

PR #66, opened 2019-02-12
Revert "pcn-router: remove arp table entries when deleting a port"

acloudiator:
Reverts polycube-network/polycube#62

As per the description, only `pcn-router` related things must have been changed, but there's a slide in edit for AUTHORS file.
Re-create the PR after review and approval from @yslu and @frisso

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -2,7 +2,6 @@ The following people, in alphabetical order, have either authored or signed
 off on commits in the Polycube repository:
 
   Aasif Shaikh                            aasif@shaikh.cc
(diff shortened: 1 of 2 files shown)


acloudiator:
> I agree with the first change (move Francesco in the contributors), but I disagree on the second change, which removes the patch form the router.
> I would leave file "ports.cpp" unchanged; the bug fix should remain.

Is there a way to partially revert the commit directly from github?

acloudiator:
Not relevant anymore.

PR #72, opened 2019-02-26
fix documentation and scripts for developers 

mauriciovasquezbernal:
Few fixes for update-code-for-all-services script:

- Datamodel is stored in .../datamodel/... instead of .../resources/...
- Add missing pcn-iptables
- Fix path for transparent-helloworld


Documentation/devel/update-code-for-all-services.sh
```diff
@@ -15,12 +16,13 @@ SERVICES=(#bridge
           pbforwarder
           router
           simplebridge
-          simpleforwarder)
+          simpleforwarder
+          transparent-helloworld)
```
mbertrone: Should we exit this script if no `POLYCUBE_HOME` is set?

```suggestion

if [ -z "$POLYCUBE_HOME" ]; then
    echo "POLYCUBE_HOME is not set"
    exit 1
fi
```

mauriciovasquezbernal: I added it, thanks!

mauriciovasquezbernal:
I removed the devel folder, scripts were moved to scripts folder and the documentation file to `Documentation/codegen.rst`.

Please notice that I only updated the paths to the files, the content of the documentation itself should be updated in few days.

mbertrone:
LGTM! Thanks Mauricio

PR #75, opened 2019-03-04
pcn-lbdsr: fix ports creation

mbertrone:
```
$polycubectl lbdsr lb1 ports add frontend-port type=FRONTEND
Port frontend-port does not exist
```
port creation was failing due to a bad port access in the code.
passing directly the index of the created port to `setFrontendPort()` solves the issue.


mauriciovasquezbernal:
Hi Matteo,

Could you please provide few words describing that is the problem solved by this PR?

mbertrone:
> Hi Matteo,
> 
> Could you please provide few words describing that is the problem solved by this PR?

Sure, I've updated PR description.

PR #77, opened 2019-03-04
Fix standalone service compilation

mauriciovasquezbernal:
Fixes the compilation of standalone services.


acloudiator:
@mauriciovasquezbernal I see the label `do not merge` by you. Do you still want to keep it on hold?

mauriciovasquezbernal:
I think this is still incomplete, so hold this for now.

mauriciovasquezbernal:
This solution is just a workaround, so I prefer to close it until we implement a proper solution for this problem: https://github.com/polycube-network/polycube/issues/91.

PR #82, opened 2019-03-13
Implement base datamodel in framework

mauriciovasquezbernal:
So far the logic that handles the elements on the base datamodel was directly into each service. 
This approach presented big disadvantages:
- developers sees a very high quantity of code that is not relevant to him
- a change in the base datamodel would require to update all services

This commit moves the handling of that logic directly to the core, so a change in the base datamodel would not require to update the services.


src/polycubed/src/base_cube.cpp
```diff
@@ -363,11 +365,20 @@ LogLevel BaseCube::get_log_level() const {
   return level_;
 }
 
-json BaseCube::to_json() const {
-  json j = {
-      {"name", name_}, {"uuid", uuid_.str()}, {"service", service_name_},
-      //{"type", type_},
-  };
+void BaseCube::set_conf(const nlohmann::json &conf) {
+  if (conf.count("loglevel")) {
```
frisso: Should we use a #define for loglevel?

mauriciovasquezbernal: I don't get the point, could you please clarify?

frisso: Hello Mauricio. Just to avoid plain strings in code, if possible, e.g.,
"loglevel"
==> 
#define LOGLEVEL_STRING "loglevel"
.... if (conf.count(LOGLEVEL_STRING(( {...

mauriciovasquezbernal: Yes, that can be done, but in this case I cannot see any advantage, the "loglevel" reference is tight coupled to the code, so in the remote case we change that we should also change methods, variables and so on.

frisso: got it, thanks!

PR #89, opened 2019-03-26
implement bash autocompletion

mauriciovasquezbernal:
Implement bash completion for polycubectl.


mauriciovasquezbernal:
> I would suggest to comment a bit more about the code in CMakeLists.txt; to me it doesn't look obvious what this code is about and how it works,

Done.

mauriciovasquezbernal:
> LGTM! Thanks for this very useful feature.
> I've just added a small dependency in install script

Thanks Matteo for the review, I squashed your commit to mine, so it is clean from the beginning.

PR #96, opened 2019-04-01
fix build failure due to the following convertion:

goldenrye:
src/polycubed/src/rest_server.cpp:130:71: error: invalid conversion from ‘int (*)(int, void*)’ to ‘int (*)(int, X509_STORE_CTX*)

### DIFF
--- a/src/polycubed/src/CMakeLists.txt
+++ b/src/polycubed/src/CMakeLists.txt
@@ -99,3 +99,4 @@ install (TARGETS polycubed DESTINATION bin)
 install(CODE
 "execute_process( COMMAND /bin/sh -c \"mkdir -p /var/log/polycube && chown $ENV{USER} /var/log/polycube \")")
 
+set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fpermissive")


mauriciovasquezbernal:
Hello @goldenrye, could you please verify that you are using the right version of pistache? (It is https://github.com/oktal/pistache/)

I haven't seen that error, that looks weird as the signature of `client_verify_callback` is `int (int, void *)` [1] and that's the same that `useSSLAuth` requires [2] 

[1] https://github.com/polycube-network/polycube/blob/master/src/polycubed/src/rest_server.h#L67
[2] https://github.com/oktal/pistache/blob/6aa428674a66c1adebc4e3527d1ac1f7dd10230b/include/pistache/endpoint.h#L115.

mauriciovasquezbernal:
@goldenrye any updated on this?

mauriciovasquezbernal:
@goldenrye ping.

goldenrye:
Will verify that today.

- Jianwen

goldenrye:
The pistache version is too old, after pulling latest pistache, the issue is gone, so close the PR.

PR #103, opened 2019-04-05
improve developers documentation

mauriciovasquezbernal:
Some improvements for developers' documentation, also update it to use docker.


Documentation/developers/codegen.rst
```diff
@@ -0,0 +1,62 @@
...
+
+In order to run the image, it is needed to mount a volume into ``/polycube-base-datamodels`` where the base datamodels are located on the host (by default ``/usr/local/include/polycube/datamodel-common/``), it is also needed to mount a volume to share the input yang model and the output folder.
+
+::
+
+  export POLYCUBE_BASEMODELS=<path to base models usually /usr/local/include/polycube/datamodel-common/>
+  docker pull polycubenetwork/polycube-codegen
+   docker run -it --user `id -u` \
+    -v $POLYCUBE_BASEMODELS:/polycube-base-datamodels \
+    -v <input yang>:/input \
```
mbertrone: I would propose 
` -v <input yang folder in host>:/input \`

Documentation/developers/codegen.rst
```diff
@@ -0,0 +1,62 @@
...
+    -v <output folder in host>:/output \
+    polycubenetwork/polycube-codegen \
+    -i /input/<yang datamodel> \
+    -o /output/<output folder>
+
```
mbertrone: Since it's possible our users are not so familiar with docker syntax, I would propose to add also an explicit example like this. What do you think @mauriciovasquezbernal ?

```
export POLYCUBE_BASEMODELS="/usr/local/include/polycube/datamodel-common/"
docker pull polycubenetwork/polycube-codegen
docker run -it --user `id -u` \
  -v $POLYCUBE_BASEMODELS:/polycube-base-datamodels \
  -v $HOME/datamodels/input:/input \
  -v $HOME/datamodels/output:/output \
  polycubenetwork/polycube-codegen \
  -i /input/iptables.yang \
  -o /output/iptables_generated_code
```

Documentation/developers/codegen.rst
```diff
@@ -0,0 +1,62 @@
+Polycube Automatic Code Generation Tools
+========================================
+
+Generating a service stub
```
mbertrone: Does it make sense adding an explicit pointer to `polycube-codegen` repository?
A Docker is provided to the user, but no pointer to the repository, in case he wants to install codegen tools locally.

mauriciovasquezbernal: Yes, it does!

mauriciovasquezbernal:
Thanks Matteo for the review, I handled your comments!

PR #112, opened 2019-04-16
add service to base datamodel

mauriciovasquezbernal:
There is an ongoing work to implement a /cubes/ api that will allow to create
a list of cubes with a single request.  In order to support that operation
there should be a service field that indicates the service type of the cube.

This commit adds the service to be base datamodel and updates polycubed to
support it.

### DIFF
--- a/src/polycubed/src/base_cube.cpp
+++ b/src/polycubed/src/base_cube.cpp
@@ -383,7 +383,7 @@ nlohmann::json BaseCube::to_json() const {
 
   j["name"] = name_;
   j["uuid"] = uuid_.str();
-  j["service"] = service_name_;
+  j["service-name"] = service_name_;
   j["type"] = cube_type_to_string(type_);
   j["loglevel"] = logLevelString(level_);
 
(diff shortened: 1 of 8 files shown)


mauriciovasquezbernal:
I have to rename the `service` leaf into `service-name` because some services already have a `service` element in their datamodel.

src/polycubed/src/base_cube.cpp
```diff
@@ -383,7 +383,7 @@ nlohmann::json BaseCube::to_json() const {
 
   j["name"] = name_;
   j["uuid"] = uuid_.str();
-  j["service"] = service_name_;
+  j["service-name"] = service_name_;
```
acloudiator: Can we use "underscore" instead of "hyphen" in `service-name` to keep the uniformity intact?

mauriciovasquezbernal: `service-name` is a json field exposed to the user while `service_name` is a variable.
In the datamodel I also used `service-name` what I think is better for user interaction.

frisso: @acloudiator I agree with Mauricio. For users, "service-name" looks better. Internally in the framework we tend to use "_" more tna "-", hence "service_name"; so this is the reason of this small difference.

PR #122, opened 2019-05-07
Netlink change: Namespace support and event extension

francescomessina:
Socket replacement:
Replacement of the libnl library socket with a raw socket; because in this way it is possible to receive netlink notifications even from the network namespaces.

Added new netlink events:
LINK_ADDED, ROUTE_ADDED, ROUTE_DELETED, NEW_ADDRESS

Removed netlink event ALL (notify_all):
Having a notification without a type is considered not very useful

### DIFF
--- a/src/polycubed/src/netlink.cpp
+++ b/src/polycubed/src/netlink.cpp
@@ -20,7 +20,15 @@
 #include <linux/if.h>
 #include <iostream>
 
+#include <arpa/inet.h>
+#include <sstream>
+#include <sys/socket.h>
+
 #include "exceptions.h"
+
+#ifndef SOL_NETLINK
+#define SOL_NETLINK 270
+#endif
 
 namespace polycube {
 namespace polycubed {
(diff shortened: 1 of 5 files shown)


mauriciovasquezbernal:
Hello Francesco.

I am getting the following compilation error in my machine:
```
/home/mvb/polycube/src/polycubed/src/netlink.cpp: In constructor ‘polycube::polycubed::Netlink::NetlinkNotification::NetlinkNotification(polycube::polycubed::Netlink*)’:
/home/mvb/polycube/src/polycubed/src/netlink.cpp:49:26: error: ‘SOL_NETLINK’ was not declared in this scope
     if (setsockopt(sock, SOL_NETLINK, NETLINK_LISTEN_ALL_NSID,(char*)&option,sizeof(option)) < 0) {
```

Could you please check it?

I also noticed that pcn-iptables uses the ALL event type of the notification [1], @mbertrone could you please confirm how this event is used there?

Thanks

[1] https://github.com/polycube-network/polycube/blob/master/src/services/pcn-iptables/src/modules/ChainSelector.cpp#L34

mbertrone:
pcn-iptables should be notified when
1) an interface is added/removed in the root namespace
2) an IP address is changed

pcn-iptables receives notifications for such events (ALL) but uses the notification just as a trigger.
A control plane thread is in charge of querying the system in order to get possible change respect to the previous state.

I'm ok by removing ALL event type if other event types will be able to support case 1 and 2.

PR #128, opened 2019-05-17
improve firewall - rules management

mbertrone:
I want to start merging some improvements to firewall service.

- Fixes for rules and counters indexes
- Minor fixes
- Porting firewall to new codegen
- Add `insert` and `delete` actions

Starting from this PR rules are handled in a contiguous way.
That means `insert`, `add`, `append`, are only allowed in the interval [0, num_rules].
`insert` or `delete` operations imply a resize and translation of rules indexes.


src/services/pcn-firewall/src/Chain.cpp
```diff
@@ -247,7 +235,7 @@ void Chain::updateChain() {
   std::map<int, std::vector<uint64_t>> protocols;
   std::vector<std::vector<uint64_t>> flags;
 
-  auto rules = getRealRuleList();
+  auto rules = rules_;
```
mauriciovasquezbernal: Is this copy really needed?, can be rules_ be passed to `conntrack_from_rules_to_map`?

src/services/pcn-firewall/src/Chain.cpp
```diff
@@ -547,6 +535,11 @@ std::vector<std::shared_ptr<ChainRule>> Chain::getRuleList() {
 }
 
 void Chain::addRule(const uint32_t &id, const ChainRuleJsonObject &conf) {
+
+  if (id > rules_.size()) {
+    throw new std::runtime_error("rule id not allowed");
```
mauriciovasquezbernal: I know this is fixed in a later commit, I think both commits should be squashed to this is fixed from the beginning.

src/services/pcn-firewall/datamodel/firewall.yang
```diff
@@ -230,6 +230,26 @@ module firewall {
...
 
+    action insert {
+      input {
+        leaf id {
+          type uint32;
+        }
+        uses "firewall:rule-fields";
+      }
+      output {
+        leaf id {
```
mauriciovasquezbernal: Is this output needed?

src/services/pcn-firewall/src/Chain.cpp
```diff
@@ -591,9 +591,156 @@ void Chain::delRuleList() {
...
+  }
+
+  if (id > rules_.size()) {
+    throw std::runtime_error("id not allowed");
+  }
+
+  auto newRule = std::make_shared<ChainRule>(*this, conf);
+
+  ChainStatsJsonObject confStats;
+  auto newStats = new ChainStats(*this, confStats);
```
mauriciovasquezbernal: This should be wrapped in a unique_ptr because if one of the following calls throws an error this memory will be leaked.

src/services/pcn-firewall/src/Chain.cpp
mbertrone: You are right, I fixed that.

src/services/pcn-firewall/src/Chain.cpp
mbertrone: I squashed this commit.

src/services/pcn-firewall/datamodel/firewall.yang
mbertrone: It is not mandatory, but I think it's nice to have. 
When inserting rules it's nice to have the id referring to the position of the new rule.

src/services/pcn-firewall/src/Chain.cpp
mbertrone: You are right, I wrapped it into a shared pointer for now, this should avoid possible memory leaks.

mbertrone:
Thanks for the review @mauriciovasquezbernal!
I've tried to address all your comments.

PR #132, opened 2019-05-23
Add veth library

francescomessina:
This PR adds 2 new libraries:
1) A veth library, able to do:
- create veth
- remove veth
- move veth into a namespace
- change the MAC address of the veth (inside and outside the namespace)
- set/unset the IP address of the veth (inside and outside the namespace)
The library uses netlink messages and the libnl library

2) A namespace library, able to perform different operation with namespaces 
as create, delete, open and run code on them.


frisso:
Francesco, I would suggest to make the suggested changes, then invite Mauricio as reviewer as well. Thanks!

mauriciovasquezbernal:
I think we should go back a step, let's start from the API of this library, then we can look at the implementation.

According to my understanding the main purpose is to handle veth pairs, create, delete and configure them.
I think this is a good idea to have a `Veth` class that is a representation of a veth pair.  Also a `VethPeer` subclass could help, let's start from it:

```
VethPeer:

void setStatus(Status);
Status getStatus();

void setIP(IP);
IP getIP();

void setMAC(MAC);
MAC getMAC();

void setNamespace(Ns);
Ns getNamespace();

string getName();

VethPeer &getPeer();

some internall stuff:
name, namespace, maybe index to speed up netlink operations?

Constructor? Destructor?
```

Tthe API of the veth should be something like

```
static Veth create(); // create a pair, interface names are random
static Veth create(name_a, name_b);  // user can specify name of interfaces
static Veth open(name_a, name_b); // just in case the user wants to handle an existing pair

void remove();

Internally:
VethPeer A_, B_; //this is composed by two peers

VethPeer & A(); // returns reference to first peer
VethPeer & B(); // same for second
```

Let's consider this example from the user point of view:

```
// create a veth pair
Veth my_pair = Veth::create("veth_in_root", "veth_in_another_ns");

// move one end to another ns (already existing)
my_pair.B().setNs("another_ns");

// set ips
my_pair.A().setIP("192.168.1.1/24"); //TODO: I am considering the IP as a string, maybe a more specific data type should be used
my_pair.B().setIP("192.168.1.2/24"); 
my_pair.A().setStatus(Status::UP);
my_pair.B().setStatus(Status::UP);

my_pair.remove();

```

Most of the methods in `VethPeer` are generic for any type of interface, so probably a base class that represents any type of interface could be implemented as well, so the user can also handle them.

@francescomessina what do you think?

mauriciovasquezbernal:
Btw, I pushed a library to handle network namespaces in the https://github.com/polycube-network/polycube/tree/add_ns_library branch, if you want to can rebase your branch to that so we introduce both features in this PR.

francescomessina:
@mauriciovasquezbernal I think your proposal it's a good idea, it would be even more understandable. I'll start work tomorrow.
For the branch  “add_ns_library” I will rebase tomorrow, of course. Thanks

src/polycubed/src/utils/veth_peer.h
```diff
@@ -0,0 +1,70 @@
...
+#include <sys/stat.h>
+#include <fcntl.h>
+#include <unistd.h>
+#include <net/if.h>
+#include <arpa/inet.h>
+#include <netinet/ether.h>
+#include <netlink/socket.h>
+#include <netlink/netlink.h>
+#include <netlink/route/link.h>
+#include <netlink/route/link/veth.h>
```
mauriciovasquezbernal: Please put here only the defines needed in the header, other should go in the cpp file.

src/polycubed/src/utils/veth_peer.h
```diff
@@ -0,0 +1,70 @@
...
+struct nl_req {
+	struct nlmsghdr nlmsg;
+	struct ifinfomsg ifinfomsg;
+};
+
+struct nl_ipreq {
+	struct nlmsghdr nlmsg;
+	struct ifaddrmsg ifaddrmsg;
+};
+
```
mauriciovasquezbernal: Are those structure  and that define needed in this header?, I think those can be moved to the cpp.

src/polycubed/src/utils/veth_peer.h
```diff
@@ -0,0 +1,70 @@
...
+    ~VethPeer();
+    void resetPeer();
+
+    void setName(const std::string &name);
+    void setStatus(const bool status);
+    void setNamespace(const std::string &ns);
+    void setMAC(const std::string &mac);
+    void setIP(const std::string &IPaddr);
+
+    std::string getName();
```
mauriciovasquezbernal: Those getters should be marked as `const`.

src/polycubed/src/utils/veth_peer.h
```diff
@@ -0,0 +1,70 @@
...
+    std::string getMAC();
+    std::string getIP();
+
+  private:
+    std::string name_;
+    bool status_;
+    std::string ns_;
+    std::string mac_;
+    std::string ip_;
+    int prefix_;
```
mauriciovasquezbernal: is this really a good idea to save all the info in the object?, I think only name could be saved here and other parameters could be queried to the kernel.

This because the interface could be modified from the external to this would not reflect that change.

src/polycubed/src/utils/veth_peer.h
```diff
@@ -0,0 +1,70 @@
...
+void* netlink_malloc(size_t size);
+struct nlmsghdr* netlink_alloc();
+int netlink_nl_send(struct nlmsghdr *nlmsg);
+struct nlmsghdr* netlink_ip_alloc();
+
+int netns_get_fd(const std::string &ns);
+
+void set_status(const std::string &iface, const std::string &ns, const bool status);
+void set_mac(const std::string &iface, const std::string &ns, const std::string &mac);
+void set_ip(const std::string &iface, const std::string &ns, const std::string &ip, int prefix);
```
mauriciovasquezbernal: Those look like internal functions to me, should they be private members?

src/polycubed/src/utils/veth_peer.cpp
```diff
@@ -0,0 +1,341 @@
...
+  prefix_ = -1;
+}
+
+/* set functions */
+void VethPeer::setName(const std::string &name) {
+  name_ = name;
+}
+
+void VethPeer::setStatus(const bool status) {
+  /* status false = DOWN; status true = UP; */
```
mauriciovasquezbernal: A bool is not the best choice here, I'd create a proper enum with DOWN and UP values.

src/polycubed/src/utils/veth_peer.cpp
```diff
@@ -0,0 +1,341 @@
...
+  if (status_ == status) {
+    return;
+  }
+  if (ns_ != "root") {
```
mauriciovasquezbernal: Using "root" to check if this is the root namespace is wrong, it is possible for a namespace to be called "root".

The can say that if ns == "" then the interface is in the root namespace.

src/polycubed/src/utils/veth_peer.cpp
```diff
@@ -0,0 +1,341 @@
...
+    std::thread t1 (set_status, name_, ns_, status);
```
mauriciovasquezbernal: Why not to use the namespace library I implemented?

src/polycubed/src/utils/veth_peer.cpp
```diff
@@ -0,0 +1,341 @@
...
+// END CLASS
+
+
+void* netlink_malloc(size_t size) {
+  void *ptr = malloc(size);
+  if (ptr == NULL) {
+    return NULL;
+  }
+  return ptr;
+}
```
mauriciovasquezbernal: Do you think that `netlink_malloc` is doing something useful?

francescomessina:
@mauriciovasquezbernal I followed all your hints, next week you tell me what you think

src/polycubed/src/utils/veth.h
```diff
@@ -0,0 +1,61 @@
...
+};
+
+class Veth {
+  public:
+    Veth(const std::string &peerA, const std::string &peerB);
+    ~Veth();
+
+    void delete_veth();
+
+    VethPeer getPeerA() const;
```
mauriciovasquezbernal: This should return a reference.

src/polycubed/src/utils/veth.cpp
```diff
@@ -0,0 +1,156 @@
...
+  } else {
+    Netlink::getInstance().set_iface_ip(name_, ip, prefix);
+  }
+}
+
+/* get functions */
+std::string VethPeer::getName() const {
+  return name_;
+}
+Namespace VethPeer::getNamespace() const {
```
mauriciovasquezbernal: `setNamespace` receives a string, I think this should return also a string.

src/polycubed/src/utils/veth.cpp
```diff
@@ -0,0 +1,156 @@
...
+void VethPeer::setNamespace(const std::string &ns) {
+  Namespace namespace_ = Namespace::open(ns);
+  int fd = namespace_.get_fd();
+  Netlink::getInstance().move_iface_into_ns(name_, fd);
+
+  ns_ = ns;
+}
+
+void VethPeer::setStatus(STATUS_IFACE status) {
+  if (ns_ != "") {
```
mauriciovasquezbernal: As a hint, you can use the `empty()` method in to do it faster.

src/polycubed/src/utils/netlink.cpp
```diff
@@ -616,5 +632,212 @@ void Netlink::detach_from_xdp(const std::string &iface, int attach_flags) {
...
+  if (!link) {
+    nl_close(sk);
+    logger->error("set_iface_mac: Invalid link");
+    return;
+  }
+
+  rtnl_link_set_name(link, iface.c_str());
+
+  struct nl_addr* addr;
+  addr = nl_addr_build(AF_LLC, ether_aton(mac.c_str()), ETH_ALEN);
```
mauriciovasquezbernal: Error check is missing here, what if mac has a wrong format?

src/polycubed/src/utils/netlink.cpp
```diff
@@ -616,5 +632,212 @@ void Netlink::detach_from_xdp(const std::string &iface, int attach_flags) {
...
+
+void Netlink::set_iface_ip(const std::string &iface, const std::string &ip, int prefix) {
+  struct nlmsghdr *nlmsg = netlink_ip_alloc();
+  struct nl_ipreq *uni = (struct nl_ipreq *)nlmsg;
+  struct rtattr *rta;
+  struct in_addr ia;
+
+  int index = get_iface_index(iface);
+  if (index == -1) {
+    logger->error("set_iface_ip: iface {0} does not exist", iface);
```
mauriciovasquezbernal: How can the caller know if there was an error? 
It is better to throw here, the caller can log the error if he wants.

src/polycubed/src/utils/netlink.h
```diff
@@ -44,6 +44,11 @@ enum class ATTACH_MODE {
   EGRESS,
 };
 
+enum class STATUS_IFACE {
```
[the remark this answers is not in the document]
francescomessina: you told me some comments above "A bool is not the best choice here, I'd create a proper enum with DOWN and UP values." 
it's wrong?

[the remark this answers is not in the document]
mauriciovasquezbernal: It's not wrong, I was asking, what about renaming to `IFACE_STATUS`?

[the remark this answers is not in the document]
francescomessina: I don't know, what name would be good? only `STATUS` is fine?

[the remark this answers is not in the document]
mauriciovasquezbernal: `STATUS_IFACE` -> `IFACE_STATUS`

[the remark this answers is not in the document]
francescomessina: sorry @mauriciovasquezbernal I didn't understand what you wanted to tell me ahahah now I'm here

mauriciovasquezbernal:
I tested the library and it is working fine.

Instead of having a Veth public constructor, it could be better to have a factory `Veth::create()`, very similar to the pattern used in the namespace library, `delete_veth()` could become `remove()`.

Last thing, in polycubed we use the snake case, `set_name`, `get_name` and so on, please stick to it.

PR #137, opened 2019-05-31
pcn-nat: fix natting-tables show

mbertrone:
ensure st_k and st_v structs are aligned, using `__attribute__((packed))`.

### DIFF
--- a/src/services/pcn-nat/src/Nat_dp_common.c
+++ b/src/services/pcn-nat/src/Nat_dp_common.c
@@ -40,6 +40,14 @@
   (sizeof(struct eth_hdr) + sizeof(struct iphdr) + \
    offsetof(struct icmphdr, checksum))
 #define IS_PSEUDO 0x10
+
+/* __attribute__((packed))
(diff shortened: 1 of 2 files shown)


src/services/pcn-nat/src/Nat_dp_common.c
```diff
@@ -52,9 +52,9 @@ struct st_k {
   uint16_t src_port;
   uint16_t dst_port;
   uint8_t proto;
-};
+} __attribute__((packed));
```
frisso: Matteo, can you please add a brief comment in the code stating why this attribute is needed?
E.g.  /* attribute packet forces alignment for this structure; otherwise segmentation fault may happen when the userspace component tries to get access to that above data */

mbertrone: Sure, comments added.

mbertrone:
@mauriciovasquezbernal could you please check the machine is building docker images?
I'm able to build docker image locally. Thanks!

PR #141, opened 2019-06-06
load old topology at startup and save each modification to file

richiMarchi:
This feature allows to load a topology from file at startup, either from the default folder or from a user selected file (ex. `polycubed --cubes-dump ~/Desktop/cubes.yaml`), keeping the whole configuration in memory.
After each modification in any of the topology of the cube, the configuration in memory is updated and another thread saves it to file updating it too.
It is possible to start the daemon with an empty topology by using the flag `--cubes-init`: if the file contains a topology, at the first update it is overwritten.
It is possible to start the daemon without the dumping to file functionality by using the flag `--cubes-nodump`.

### DIFF
--- a/src/polycubed/src/config.cpp
+++ b/src/polycubed/src/config.cpp
@@ -32,13 +32,17 @@ namespace configuration {
 
 #define LOGLEVEL spdlog::level::level_enum::info
 #define DAEMON false
+#define CUBESINIT false
+#define CUBESNODUMP false
 #define PIDFILE "/var/run/polycube.pid"
 #define SERVER_PORT 9000
 #define SERVER_IP "localhost"
 #define LOGFILE "/var/log/polycube/polycubed.log"
 #define CONFIGFILEDIR "/etc/polycube"
 #define CONFIGFILENAME "polycubed.conf"
+#define CUBESDUMPFILE "cubes.yaml"
 #define CONFIGFILE (CONFIGFILEDIR "/" CONFIGFILENAME)
+#define CUBESDUMPFILEPATH (CONFIGFILEDIR "/" CUBESDUMPFILE)
 
 static void show_usage(const std::string &name) {
   std::cout << std::boolalpha;
@@ -64,6 +68,11 @@ static void show_usage(const std::string &name) {
             << ")" << std::endl;
   std::cout << "--configfile: configuration file (default: " << CONFIGFILE
             << ")" << std::endl;
+  std::cout << "--cubes-dump: file that keeps the last topology, "
+               "including the configuration of all cubes (default: " << CUBESDUMPFILEPATH
+            << ")" << std::endl;
+  std::cout << "--cubes-init: starts the daemon with an empty topology" << std::endl;
+  std::cout << "--cubes-nodump: starts the daemon without dumping updates to file" << std::endl;
   std::cout << "--cert-blacklist: path to black listed certificates"
             << std::endl;
   std::cout << "--cert-whitelist: path to white listed certificates"
@@ -95,7 +104,10 @@ Config::Config()
       server_port(SERVER_PORT),
       server_ip(SERVER_IP),
       logfile(LOGFILE),
-      configfile(CONFIGFILE) {}
+      configfile(CONFIGFILE),
+      cubes_dump_file(CUBESDUMPFILEPATH),
+      cubes_init(CUBESINIT),
+      cubes_nodump(CUBESNODUMP){}
 
 Config::~Config() {}
 
@@ -180,6 +192,31 @@ void Config::setLogFile(const std::string &value) {
   logfile = value;
 }
 
+std::string Config::getCubesDumpFilePath() const {
+  return cubes_dump_file;
+}
+
+void Config::setCubesDumpFilePath(const std::string &value) {
+  CHECK_OVERWRITE("cubes-dump", value, cubes_dump_file, CUBESDUMPFILEPATH);
+  cubes_dump_file = value;
+}
+
+bool Config::getCubesInitTopology() const {
+  return cubes_init;
+}
+
+void Config::setCubesInitTopology() {
+  cubes_init = true;
+}
+
+bool Config::getCubesNoDump() const {
+  return cubes_nodump;
+}
+
+void Config::setCubesNoDump() {
+  cubes_nodump = true;
+}
+
 std::string Config::getCertPath() const {
   return cert_path;
 }
@@ -246,6 +283,8 @@ void Config::create_configuration_file(const std::string &path) {
   file << "addr: " << server_ip << std::endl;
   file << "# file to save polycube logs" << std::endl;
   file << "logfile: " << logfile << std::endl;
+  file << "# file to save last topology" << std::endl;
+  file << "cubes-dump: " << cubes_dump_file << std::endl;
   file << "# Security related:" << std::endl;
   file << "# server certificate " << std::endl;
   file << "#cert: path_to_certificate_file" << std::endl;
@@ -270,6 +309,9 @@ void Config::dump() {
   logger->info(" port: {}", server_port);
   logger->info(" addr: {}", server_ip);
   logger->info(" logfile: {}", logfile);
+  logger->info(" cubes-dump: {}", cubes_dump_file);
+  logger->info(" cubes-init: {}", cubes_init);
+  logger->info(" cubes-nodump: {}", cubes_nodump);
   if (!cert_path.empty()) {
     logger->info(" cert: {}", cert_path);
   }
@@ -363,6 +405,9 @@ static struct option options[] = {
     {"cacert", required_argument, NULL, 5},
     {"cert-whitelist", required_argument, NULL, 6},
     {"cert-blacklist", required_argument, NULL, 7},
+    {"cubes-dump", required_argument, NULL, 8},
+    {"cubes-init", no_argument, NULL, 9},
+    {"cubes-nodump", no_argument, NULL, 10},
     {NULL, 0, NULL, 0},
 };
 
@@ -407,6 +452,15 @@ void Config::load_from_cli(int argc, char *argv[]) {
       break;
     case 7:
       setCertBlacklistPath(optarg);
+      break;
+    case 8:
+      setCubesDumpFilePath(optarg);
+      break;
+    case 9:
+      setCubesInitTopology();
+      break;
+    case 10:
+      setCubesNoDump();
       break;
     }
   }
--- a/src/polycubed/src/cubes_dump.cpp
+++ b/src/polycubed/src/cubes_dump.cpp
@@ -1,0 +1,411 @@
+/*
+ * Copyright 2019 The Polycube Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <fstream>
+#include "cubes_dump.h"
+#include "config.h"
+#include "rest_server.h"
+#include "server/Resources/Body/ListKey.h"
+#include "server/Resources/Endpoint/Resource.h"
+
+namespace polycube::polycubed {
+
+using Rest::Resources::Body::ListKey;
+using Rest::Resources::Endpoint::Operation;
+
+CubesDump::CubesDump() : dump_enabled_(false), killSavingThread_(false) {
+  save_in_file_thread_ = std::make_unique<std::thread>(&CubesDump::SaveToFile, this, configuration::config.getCubesDumpFilePath());
+}
+
+CubesDump::~CubesDump() {
+  killSavingThread_ = true;
+  waitForUpdate_.notify_one();
+  save_in_file_thread_->join();
+}
+
+void CubesDump::Enable() {
+  dump_enabled_ = true;
+}
+
+void CubesDump::UpdateCubesConfig(const std::string& resource,
+                                  const nlohmann::json& body,
+                                  const ListKeyValues &keys,
+                                  Rest::Resources::Endpoint::Operation opType,
+                                  Rest::Resources::Endpoint::ResourceType resType) {
+
+  // TODO: remove this check
+  if (configuration::config.getCubesNoDump()) {
+    return;
+  }
+
+  std::string resourcePath = resource.substr(RestServer::base.size(), resource.size());
+
+  std::stringstream ssResource(resourcePath);
+  std::vector<std::string> resItem;
+  std::string tokenResource;
+
+  while (std::getline(ssResource, tokenResource, '/')) { /* Get path elements */
+    resItem.push_back(tokenResource);
+  }
+
+  /*
+   * Depending on the operation, we look into the resource string, the body
+   * and the ListKeyValues to update the configuration of the cubes in the map
+   * (<cubeName, cubeConfiguration>) we have in memory.
+   * If it is not the initial topology load, the whole configuration is saved
+   * to file after each modification by a separate thread, to avoid to overload
+   * the server thread.
+   */
+  std::lock_guard<std::mutex> guardConfigMutex(cubesConfigMutex_);
+  switch (opType) {
+  case Operation::kCreate:
+  case Operation::kReplace:
+    UpdateCubesConfigCreateReplace(resItem, body, keys, resType);
+    break;
+
+  case Operation::kUpdate:
+    UpdateCubesConfigUpdate(resItem, body, keys, resType);
+    break;
+
+  case Operation::kDelete:
+    UpdateCubesConfigDelete(resItem, body, keys, resType);
+    break;
+  }
+
+  /*
+   * if the dump is enabled, notify the saving thread that an update
+   * to the topology occured and needs to be dumped to file
+   */
+  if (dump_enabled_) {
+    listOfPendingChanges_++;
+    waitForUpdate_.notify_one();
+  }
+}
+
+/*
+ * This updates the configuration when the connect or disconnect special commands
+ * are used, setting another cube's port as peer for a certain cube.
+ */
+void CubesDump::UpdatePortPeer(const std::string& cubeName, const std::string& cubePort, std::string peer) {
+
+  // TODO: remove this check
+  if (configuration::config.getCubesNoDump()) {
+    return;
+  }
+
+  std::lock_guard<std::mutex> guardConfigMutex(cubesConfigMutex_);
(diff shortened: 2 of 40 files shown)


src/polycubed/src/polycubed.cpp
```diff
@@ -278,6 +278,9 @@ int main(int argc, char *argv[]) {
   // register services that are shipped with polycube
   load_services(*core);
 
+  // load the last topology created
+  restserver->load_last_topology();
```
frisso: Given the previous comment, if the user asked to start with an empty topology, we don't have to load the previous topology here.

src/polycubed/src/rest_server.cpp
```diff
@@ -26,12 +27,14 @@
...
 #include "server/Resources/Data/AbstractFactory.h"
 #include "server/Server/ResponseGenerator.h"
+#include "config.h"
 
 namespace polycube {
 namespace polycubed {
 
 std::string RestServer::whitelist_cert_path;
 std::string RestServer::blacklist_cert_path;
+bool RestServer::startup;
```
frisso: Add a comment or rename the variable. It does not sound obvious to me what a boolean "startup" means.

src/polycubed/src/rest_server.cpp
```diff
@@ -150,6 +157,27 @@ void RestServer::shutdown() {
   } catch (const std::runtime_error &e) {
     logger->error("{0}", e.what());
   }
+  Rest::Resources::Endpoint::Resource::kill = true;
+  Rest::Resources::Endpoint::Resource::data_cond.notify_one();
+}
+
+void RestServer::load_last_topology() {
+  startup = true;
+  std::ifstream myFile(get_last_topology_path());
```
frisso: Use more meaningful names. MyFile does not mean nothing.
topologyFile looks much better than myFile.

src/polycubed/src/rest_server.cpp
```diff
@@ -150,6 +157,27 @@ void RestServer::shutdown() {
...
+  if (myFile.is_open()) {
+    std::stringstream buffer;
+    buffer << myFile.rdbuf();
+    myFile.close();
+    json j = json::parse(buffer.str());
+    logJson(j);
+    for (auto &it : j) {
+      core.get_service_controller(it["service-name"]).get_management_interface()->get_service()
```
frisso: Add a comment: what are you going to do here?

src/polycubed/src/server/Resources/Endpoint/LeafResource.cpp
```diff
@@ -113,6 +113,13 @@ void LeafResource::CreateReplaceUpdate(const Pistache::Rest::Request &request,
     auto op = Operation::kUpdate;
     auto resp = WriteValue(cube_name, jbody, keys, op);
     errors.push_back(resp);
+    if (resp.error_tag == ErrorTag::kOk ||
```
frisso: Not clear to me what you're doing. Please add a comment.
The same for L178 below.

src/polycubed/src/server/Resources/Endpoint/Resource.cpp
```diff
@@ -1,3 +1,7 @@
+#include <utility>
```
frisso: Include files should be placed after the copyright banner.

src/polycubed/src/server/Resources/Endpoint/Resource.cpp
```diff
@@ -33,4 +46,152 @@ Operation Resource::OperationType(bool update, bool initialization) {
...
+  std::string token;
+
+  while(std::getline(ss, token, '/')) {
+    resItem.push_back(token);
+  }
+
+  const std::string &serviceName = resItem[0];
+  const std::string &cubeName = resItem[1];
+
+  std::lock_guard<std::mutex> lg(mutex);
```
frisso: What does it mean "lg"? Use more meaningful names.

src/polycubed/src/server/Resources/Endpoint/Resource.cpp
```diff
@@ -33,4 +46,152 @@ Operation Resource::OperationType(bool update, bool initialization) {
...
+  switch (opType) {
```
frisso: Add some comments presenting how the system works.
If I remember correctly, configuration is saved in memory (a map?), then periodically saved on disk by the detached thread. This must be documented at least in the code, otherwise it becomes difficult to follow.

src/polycubed/src/server/Resources/Endpoint/Resource.cpp
```diff
@@ -33,4 +46,152 @@ Operation Resource::OperationType(bool update, bool initialization) {
...
+        }
+        if (!deleted) {
+          item->erase(resItem[resItem.size() - 1]);
+        }
+      }
+      break;
+    }
+  }
+
+  if (!RestServer::startup) {
```
frisso: Why is this code needed? Add a comment.

src/polycubed/src/server/Resources/Endpoint/Resource.cpp
```diff
@@ -33,4 +46,152 @@ Operation Resource::OperationType(bool update, bool initialization) {
...
+    toSave++;
+    data_cond.notify_one();
+  }
+}
+
+void Resource::SaveToFile(const std::string& path) {
+  while (true) {
+    std::unique_lock<std::mutex> uniqueLock(mutex);
```
frisso: Change name.
You don't have to call "lock" a variable that is a lock. It means nothing.
You have to tell people what is this lock doing: what is it protecting? E.g. dumpfile_lock looks more understandable. Adding a comment when you define the variable looks even better.

src/polycubed/src/config.cpp
```diff
@@ -64,6 +66,8 @@ static void show_usage(const std::string &name) {
             << ")" << std::endl;
   std::cout << "--configfile: configuration file (default: " << CONFIGFILE
             << ")" << std::endl;
+  std::cout << "--cubes-file: file to save last topology (default: " << LASTTOPOLOGYPATH
```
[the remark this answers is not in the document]
mauriciovasquezbernal: What should so `--cubes-init`, create a new and empty cubes file?
Is this option really needed?, can't the user just remove the cubes file before running polyucubed?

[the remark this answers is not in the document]
frisso: Clearing up the file does not look intuitive to me. I expect something like "start with a void config", which is exactly what I was suggesting. Open to suggestion for this new command line option.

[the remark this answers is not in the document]
mauriciovasquezbernal: I don't know if that feature is really useful, anyway, I am not against it.

I have however a doubt about its behavior, What should happen if there is an already existing cubes file? Should it be cleaned up and then used to save the new topology?

Should we also provide an option to disable the saving of the cubes file? Just in case somebody doesn't need it and wants to avoid it's penalty.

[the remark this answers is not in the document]
richiMarchi: At the moment I implemented it (not pushed yet) without cleaning the file up. At the first action (for instance a creation of a cube) the file is overwritten, otherwise it remains with the previous configuration.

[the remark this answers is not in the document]
frisso: > What should happen if there is an already existing cubes file? Should it be cleaned up and then used to save the new topology?

==> If you keep the state, the cubes.yaml file will be reinitialized and used to store the new topology (and config), i.e., the current state.


> Should we also provide an option to disable the saving of the cubes file? Just in case somebody doesn't need it and wants to avoid its penalty.

==> Very good point. I think "--cubes-nodump" can be a very good option then.
Thanks!

mauriciovasquezbernal:
Hello @richiMarchi, I didn't go through the whole code yet, I tested and found a problem, consider the following list of commands:

```
polycubectl helloworld add hw0
polycubectl hw0 ports add p1
polycubectl router r0 
polycubectl r0 ports add p1 ip=192.168.0.1 netmask=255.255.255.0
polycubectl hw0 ports p1 set peer=r0:p1
polycubectl r0 ports p1 set peer=hw0:p1
```

It produces the following cubes file:

```
$ sudo cat /etc/polycube/cubes.yaml
[
  {
    "service-name": "helloworld",
    "name": "hw0",
    "type": "TC",
    "loglevel": "INFO",
    "action": "DROP",
    "ports": [
      {
        "peer": "r0:p1",
        "name": "p1"
      }
    ]
  },
  {
    "service-name": "router",
    "name": "r0",
    "type": "TC",
    "loglevel": "INFO",
    "ports": [
      {
        "ip": "192.168.0.1",
        "netmask": "hw0:p1",
        "name": "p1"
      }
    ]
  }
]
```

You can notice that the peer has been saved in the netmask field of port1 in r0.

mauriciovasquezbernal:
There are other two problems that we'll have to solve after the base code is ready.

The first one is that `connect`, `disconnect`, `attach` and `detach` commands aren't processed by that part of the code, so changes made by those are not saved to the file.

The second problem is that if a cube's port has a peer set to another cube's port it'll fail at creation because that cube does not exist yet, in other words, a cubes file like the following fails:

```
sudo cat /etc/polycube/cubes.yaml
[
  {
    "service-name": "helloworld",
    "name": "hw0",
    "type": "TC",
    "loglevel": "INFO",
    "action": "DROP",
    "ports": [
      {
        "peer": "hw1:p1",
        "name": "p1"
      }
    ]
  },
  {
    "service-name": "helloworld",
    "name": "hw1",
    "type": "TC",
    "loglevel": "INFO",
    "action": "DROP",
    "ports": [
      {
        "peer": "hw0:p1",
        "name": "p1"
      }
    ]
  }
]
```

I think we'll have to relax the requirement that a port must exist when setting the peer.

src/polycubed/src/server/Resources/Endpoint/Resource.cpp
```diff
@@ -33,4 +42,193 @@ Operation Resource::OperationType(bool update, bool initialization) {
...
+    if (toSave.load() == 0 && !kill) {
+      waitForUpdate.wait(cubesConfigLock);
+    }
+    if (toSave.load() == 0 && kill) {
+      break;
+    }
+    std::map<std::string, nlohmann::json> copyConfig(cubesConfig);
+    toSave.store(0);
+    cubesConfigLock.unlock();
+    std::ofstream myFile(path);
```
frisso: Rename myFile in something more meaningful.

src/polycubed/src/rest_server.cpp
frisso: Please add the following comment in the code:
// This boolean is true when the rest server starts. This will be used not to save the current topology at startup.
// This flag is then reset to false, and all the incoming configuration commands will be saved in the dump file.

src/polycubed/src/CubesDump.cpp
```diff
@@ -0,0 +1,293 @@
...
+  // map saving key values corresponding to proper list name
+  std::map<std::string, std::vector<polycube::polycubed::Rest::Resources::Body::ListKeyValue>> keyValues;
+  for (auto &key : keys) {
+    keyValues[key.list_element].push_back(key);
+  }
+
+  switch (resType) {
+    case polycube::polycubed::Rest::Resources::Endpoint::ResourceType::Service: {
+      nlohmann::json serviceField = nlohmann::json::object();
+      serviceField["service-name"] = serviceName;
```
mauriciovasquezbernal: Is this really needed?, Isn't `service-name` already present in the json object?

src/polycubed/src/polycubed.cpp
```diff
@@ -278,6 +279,13 @@ int main(int argc, char *argv[]) {
   // register services that are shipped with polycube
   load_services(*core);
 
+  // load the last topology created
+  if (!config.getCubesInitTopology()) {
+    restserver->load_last_topology();
+  }
+  //Start saving thread
+  std::thread(CubesDump::SaveToFile, config.getCubesDumpFilePath()).detach();
```
mauriciovasquezbernal: if `cubes-no-dump` is set this thread should not be created.

src/polycubed/src/cubes_dump.cpp
```diff
@@ -0,0 +1,336 @@
...
+  }
+
+  // if startup is true, then it is the first load and the save thread must not be notified
+  if (enabled_) {
+    toSave_++;
+    waitForUpdate_.notify_one();
+  }
+}
+
+void CubesDump::UpdatePortPeer(const std::string& cubeName, const std::string& cubePort, std::string peer) {
```
frisso: Not clear what this function is currently doing.
Please add a comment.

src/polycubed/src/cubes_dump.cpp
```diff
@@ -0,0 +1,336 @@
...
+  }
+
+  // if startup is true, then it is the first load and the save thread must not be notified
+  if (enabled_) {
+    toSave_++;
+    waitForUpdate_.notify_one();
+  }
+}
+
+void CubesDump::UpdateCubesConfigCreateReplace(const std::vector<std::string> &resItem,
```
frisso: Add a comment: what is this function about?
In general, functions should have a short comment at the beginning, explaining what they are doing.
Please extend the comment to the other functions as well.

src/polycubed/src/cubes_dump.cpp
```diff
@@ -0,0 +1,336 @@
...
+  }
+}
+
+void CubesDump::SaveToFile(const std::string& path) {
+  while (true) {
+    // mutex with condition variable waitForUpdate on cubesConfig_
+    std::unique_lock<std::mutex> cubesConfigLock(cubesConfigMutex_);
+    // if there are no updates from last write to file,
+    // either wait for updates (kill=false) or exit if the daemon is shutting down (kill=true)
+    if (toSave_.load() == 0 && !kill_) {
```
frisso: 1) Not clear to me what 'toSave' means.
Is something like "listOfPendingChanges"? In general, select meaningful names.
2) What does "toSave_.load()" means? Why load? Are we loading or saving? The comment is not so helpful either.

src/polycubed/src/cubes_dump.cpp
```diff
@@ -0,0 +1,336 @@
...
+      waitForUpdate_.wait(cubesConfigLock);
+    }
+
+    if (toSave_.load() == 0 && kill_) {
+      break;
+    }
+
+    std::map<std::string, nlohmann::json> copyConfig(cubesConfig_);
+    toSave_.store(0);
```
frisso: What are we storing here? Add a comment.

src/polycubed/src/cubes_dump.h
```diff
@@ -0,0 +1,78 @@
...
+  // mutex on the access to cubesConfig
+  std::mutex cubesConfigMutex_;
+  // cubes configuration <name, configuration>
+  std::map<std::string, nlohmann::json> cubesConfig_;
+  //how many updates have been made while thread saving to file
+  std::atomic<int> toSave_;
+  // wait until an update occurs
+  std::condition_variable waitForUpdate_;
+  // the saving thread ends if the daemon is shutting down (kill=true)
+  bool kill_;
```
frisso: Should "kill" be renamed in "kill_requested"? Or "kill_pending"?

src/polycubed/src/polycubed.cpp
```diff
@@ -278,6 +284,22 @@ int main(int argc, char *argv[]) {
...
   load_services(*core);
 
+  // create a cubes dump instance if needed
+  if (!config.getCubesNoDump()) {
+    cubesdump = new CubesDump();
+    core->set_cubes_dump(cubesdump);
+  }
+
+  // load the last topology created
+  if (!config.getCubesInitTopology()) {
```
frisso: Replace the comment with
// In case the user does not want to initialize the Polycube virtual network, let's load the last topology that was present when the daemon was shut down.

src/polycubed/src/polycubed_core.cpp
```diff
@@ -243,6 +243,7 @@ bool PolycubedCore::try_to_set_peer(const std::string &peer1,
     }
     auto port = cube->get_port(match[2]);
     port->set_peer(peer2);
+    cubes_dump_->UpdatePortPeer(cube->get_name(), port->name(), peer2);
```
frisso: Add a comment why is this line needed.

src/polycubed/src/polycubed_core.h
```diff
@@ -92,6 +98,7 @@ class PolycubedCore {
   std::shared_ptr<spdlog::logger> logger;
   RestServer *rest_server_;
   BaseModel *base_model_;
+  CubesDump *cubes_dump_;
```
frisso: Add a comment about what is this member.

src/polycubed/src/port.cpp
```diff
@@ -161,6 +161,12 @@ void Port::set_conf(const nlohmann::json &conf) {
   if (conf.count("peer")) {
     set_peer(conf.at("peer").get<std::string>());
   }
+
+  if (conf.count("tcubes")) {
```
frisso: Add a comment about what we're doing here.

src/polycubed/src/rest_server.cpp
```diff
@@ -152,6 +157,57 @@ void RestServer::shutdown() {
...
+   */
+  auto peers = utils::strip_port_peers(cubes);
+  std::vector<Response> resp = {{ErrorTag::kNoContent, nullptr}};
+  bool error = false;
+
+  for (auto &cube : cubes) {
+    auto &s = core.get_service_controller(cube["service-name"]);
+    auto res = s.get_management_interface()->get_service()
+            ->CreateReplaceUpdate(cube["name"], cube, false, true);
+    if (res.size() != 1 || res[0].error_tag != kCreated) {
```
frisso: Add a space before this line.

src/polycubed/src/server/Resources/Data/BaseModel/ConcreteFactory.cpp
```diff
@@ -212,8 +214,42 @@ std::unique_ptr<Endpoint::LeafListResource> ConcreteFactory::RestLeafList(
...
+    tree_names_.pop();
+    auto leaf = tree_names_.front();
+    if (leaf == "tcubes") {
+      read_handler_ = [local_core](const std::string &cube_name,
+                                   const ListKeyValues &keys) -> Response {
+        return local_core->base_model()->get_port_tcubes(cube_name,
+                                                         keys[0].value);
+      };
+    } else {
+      throw std::runtime_error("unkown element found in base datamodel: " +
```
frisso: unkown ==> unknown

src/polycubed/src/server/Resources/Endpoint/Service.cpp
```diff
@@ -84,11 +88,33 @@ Service::CreateReplaceUpdate(const std::string &name, nlohmann::json &body, bool
       return std::move(body_errors);
     }
 
+   /*
+    * This is possible that a port inside a cube includes a list of
```
frisso: This is ==> It is

src/services/pcn-nat/src/Nat.cpp
```diff
@@ -83,6 +83,12 @@ void Nat::attach() {
     auto temp = get_parent_parameter("ip");
     external_ip_ = temp.substr(1, temp.length() - 2);  // remove qoutes
     logger()->info("external ip is : {}", external_ip_);
+    // if masquerate is enabled this is needed to update the ip address
```
frisso: masquerate ==> masquerade
this is needed ==> we need

frisso:
Really great; thanks, Riccardo!

PR #146, opened 2019-06-07
Network Monitor

mbertrone:
This PR adds a transparent service called `Network Monitor`.
The purpose of this module is collecting info over a link, that can be used by a remote backend server to detect attacks and anomalies.

Note: right now the implementation is not fully completed; the module is returning pseudo-random metrics.


acloudiator:
@mbertrone This seems to be a nice feature. Will it be applied to the framework; like all the polycube services benefit from this or only specific services? Thanks!

@frisso I recommend keeping this on hold before we merge to master, as this isn't a fully functional feature. Do you share the same thought?

frisso:
@acloudiator 
1) This is a generic service, so in principle you could plug it in any place (any netdev, and any connection between services). However, this particular service will be specialized in SYN flooding attacks, so in practice its effectiveness is limited only to that scope.
2) To facilitate the work of other partners in the ASTRID EU project, it would be nice if we can integrate it in the master. It should be a matter of weeks to have it 100% ready, dataplane included.

acloudiator:
> 2. To facilitate the work of other partners in the ASTRID EU project, it would be nice if we can integrate it in the master. It should be a matter of weeks to have it 100% ready, dataplane included.

IMHO, We may first have it working in separate branch.
Then we may validate/pass some tests before we merge into master. 

Please do let me know If you still think otherwise. Thanks!

mbertrone:
After @frisso suggestion to rename the service `synflood` detector, and in order to avoid misunderstanding on service purpose, I'm going to close this PR.
Please refer to #152

PR #147, opened 2019-06-10
pcn-firewall: implement datapath optimizations

mbertrone:
**pcn-firewall: introduce datapath pipeline early break**
this commit introduces the support for some firewall
filtering pipeline modules to early break the pipeline
if no rule is matched.

the idea is that during the execution of the pipeline
some modules are implemented in a way in wich the cost
of checking if the resulting bitvector is all zero
is almost negligible.

this support enables to speedup the execution since is no
longer needed to go through all the pipeline modules
if this condition is matched.

**pcn-firewall: dynamic pipeline reordering**
this commit introduces support for pipeline reordering.

in case some filtering pipeline blocks can early break
the pipeline, those modules are injected before others.

**pcn-firewall: update connection tracking module**
this commit introduces the same connection tracking
module used by pcn-iptables

Main features:
- avoid double lookup by ordering hashmap keys
- avoid to call bpf_ktime_get_ns() by introducing
   a customized timestamp written by control plane

**pcn-firewall: add HORUS optimization**
HORUS - Homogeneous RUleset analySis

Horus optimization allows to
 a) offload a group of contiguous rules matching on same field
 b) match the group of offloaded rules with complexity O(1) - single hashmap lookup
 c) dynamically adapting to different groups of rules, matching each
     combination of ipsrc/dst, portsrc/dst, tcpflags
 d) dynamically check when the optimization is possible according to current
     ruleset. It means check orthogonality
     of rules before the offloaded group, respect to the group itself.

each pkt received by the program, is looked-up vs the HORUS HASHMAP.
hit:
 -DROP action: drop the packet;
 -ACCEPT action: goto CTLABELING and CTTABLEUPDATE without going through pipeline
miss:
 -GOTO all pipeline steps


frisso:
@acloudiator Can you please merge this? The problem (Docker build was failing) has gone now.

acloudiator:
> @acloudiator Can you please merge this? The problem (Docker build was failing) has gone now.

@frisso Can you please be the second reviewer for this, so that this can be merged?

PR #148, opened 2019-06-10
polycubed: fix packet out deadlock when attaching transparent cubes

mauriciovasquezbernal:
A dead lock is possible when a transparent cube is added to a port
    and there is a packetout operation at the same time.
    If port1 and port2 are connected, a packetout operation is going
    on port1 and a cube is attached to port2
    
```
    Thread1:                          | Thread2:
    port1->send_packet_out()          | port2->add_cube()
      lock(port1)                     |   lock(port2)
      ...                             |   ...
      port2->get_index()              |   port1->set_next_index()
        lock(port2)                   |     lock(port1)
  ```  
    The main cause of this is that send_packet_out() uses get_index() that
    internally uses a lock.  This commit modifies get_index() so it doesn't

### DIFF
--- a/src/polycubed/src/port.cpp
+++ b/src/polycubed/src/port.cpp
@@ -49,11 +49,10 @@ uint16_t Port::index() const {
 }
 
 uint16_t Port::get_index() const {
-  std::lock_guard<std::mutex> guard(port_mutex_);
-  return __get_index();
+  return port_index_;
 }
 
-uint16_t Port::__get_index() const {
+uint16_t Port::calculate_index() const {
   // check if there is ingress-enabled transparent cube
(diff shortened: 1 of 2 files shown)


frisso:
@acloudiator Looks Docker build is ok now.

frisso:
@acloudiator Docker build has been fixed (the problem was in Docker build, not in this commit). Can you please merge in master? Thanks!

mauriciovasquezbernal:
This is not strictly necessary to protect the index in `get_index` function as this is only used by the send_packet_out function, the worst thing that can happen is that some of the packets out are not processed by the transparent cube being attached for while until the index is updated.

PR #144, opened 2019-06-07
pcn-iptables: update and unify documentation

mbertrone:
- unify documentation under /Documentation
- remove docs under /src/components/iptables

My only doubt is about keeping or not the readme file under ` src/components/iptables/README.md`

PR #149, opened 2019-06-10
elaboration to steps of pcn-iptables

acloudiator:
Few elaborations to pcn-iptables document which observed during installation.


mauriciovasquezbernal:
@acloudiator why not to integrate this into https://github.com/polycube-network/polycube/pull/144 instead of having a new PR that changes a couple of lines?

src/components/iptables/README.md
```diff
@@ -18,7 +18,7 @@ make && sudo make install
 
 ```
 # Start polycubed, in other terminal (or background)
-sudo polycubed
+sudo polycubed &
```
mbertrone: Another possibility is to launch polycubed in daemon mode.

acloudiator:
> I just had a better look at this PR. I think this is unnecessary, #144 is deleting this readme file and moving everything under Documentation.

Do you recommend to move it under Documentation? Or You find the PR **unnecessary** and drop it??

src/components/iptables/README.md
mauriciovasquezbernal: This is in fact the best way to do it.
To use polycubed --daemon.

mauriciovasquezbernal:
> > I just had a better look at this PR. I think this is unnecessary, #144 is deleting this readme file and moving everything under Documentation.
> 
> Do you recommend to move it under Documentation? Or You find the PR **unnecessary** and drop it??

I think we can move this into Documentation.

src/components/iptables/README.md
acloudiator: Wasn't the `d` at the end of polycube resembles `daemon`. I don't know what that `d` is then :)
Certainly, I'll modify this option with the neat option `--daemon`.

frisso: @mauriciovasquezbernal I agree with Aasif, I had the same issue in the past. Having 'd' at the end of 'polycubed', I also assumed it runs as a daemon. Personally, I feel more clear to reverse the behaviour, i.e., polycubed runs as daemon, while something else like 'polycubed -foreground' runs in the terminal.

mauriciovasquezbernal:
@acloudiator could you please update this so we can merge this one?

PR #150, opened 2019-06-11
update to new bcc

mauriciovasquezbernal:
The 0.7.0 version we're using is not compatible with kernels 5.0, update
this to a 0.10.0 that is compatible.

Note: this should be merged while https://github.com/polycube-network/bcc/tree/polycube is updated to https://github.com/polycube-network/bcc/tree/polycube_on_v0.10.0

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -1,5 +1,7 @@
 find_package(LibYANG REQUIRED)
 
+include_directories(${CMAKE_SOURCE_DIR}/src/libs/bcc/src/cc/libbpf/include/uapi)
+
 option(INSTALL_CLI "installs the polycube CLI" ON)
 option(HAVE_POLYCUBE_TOOLS "uses the polycube-tools package" OFF)
 option(ENABLE_PCN_IPTABLES "enables the pcn-iptables" OFF)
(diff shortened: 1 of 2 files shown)


mauriciovasquezbernal:
Please don't merge it. Tests for firewall and iptables are failing.

mauriciovasquezbernal:
> shouldn't the changes under directory src/libs/bcc submitted to iovisor/bcc upstream and then sync up with the main upstream, how we manage the difference if we check into polycube repo only?
> 

That should be the best plan, to propose our features to bcc and merge them there. However this is something we haven't done (I don't know if we will do).
I don't get your point about the difference, what do you mean?

> Does this change fixes defect in shared table implementation or a common issue for BPF module?

This change uses a recent version of bcc that works in the newer kernel versions > 5.0, functionality remains the same.

goldenrye:
Does this change back-compatible with kernel 4,x and older kernel?

mauriciovasquezbernal:
I used it in 4.19 and worked good. This should also be ok in older versions.

mauriciovasquezbernal:
Tests are now passing: http://130.192.225.104:8080/blue/organizations/jenkins/polycube-test/detail/PR-150/22/pipeline.
There is a nat test that fails randomly also in the master, so we can say this is not because of this PR.

frisso:
@acloudiator Can we merge already?

frisso:
@acloudiator Tag has been created. Can you please merge it? Thanks!

acloudiator:
> @acloudiator Tag has been created. Can you please merge it? Thanks!

The tests are still failing - please checck the recent one. (http://130.192.225.104:8080/blue/organizations/jenkins/polycube-test/activity/?branch=PR-150).

acloudiator:
Update from @frisso 

> Compilation problems are related to other issues, bugs that we already inserted in the code and that  @mauriciovasquezbernal is taking care of. Nothing related to this PR.

Based on the above confirmation and tested by @mauriciovasquezbernal; this PR can be merged.

PR #151, opened 2019-06-11
pcn-iptables: fix horus module

mbertrone:
- avoid incompatibility with conntrack rules
- increase horus max size
- use attribute packed to avoid memory alignment issues

### DIFF
--- a/src/services/pcn-iptables/src/Utils.cpp
+++ b/src/services/pcn-iptables/src/Utils.cpp
@@ -534,10 +534,15 @@ bool Chain::interfaceFromRulesToMap(
   return brk;
 }
 
-void Chain::fromRuleToHorusKeyValue(std::shared_ptr<ChainRule> rule,
+bool Chain::fromRuleToHorusKeyValue(std::shared_ptr<ChainRule> rule,
                                     struct HorusRule &key,
                                     struct HorusValue &value) {
   key.setFields = 0;
+
+  try {
+    rule->getConntrack();
+    return false;
+  } catch (std::runtime_error) {}
 
   try {
     IpAddr ips;
@@ -590,7 +595,7 @@ void Chain::fromRuleToHorusKeyValue(std::shared_ptr<ChainRule> rule,
     value.action = 1;
 
   value.ruleID = rule->getId();
-  return;
+  return true;
 }
 
 void Chain::horusFromRulesToMap(
@@ -599,7 +604,6 @@ void Chain::horusFromRulesToMap(
   struct HorusRule key;
   struct HorusValue value;
 
-  bool first_rule = true;
   uint64_t set_fields = 0;
 
   // find match pattern for first rule (e.g. 01101 means ips proto ports set)
@@ -607,22 +611,21 @@ void Chain::horusFromRulesToMap(
   int i = 0;
 
   for (auto const &rule : rules) {
-    i++;
-    fromRuleToHorusKeyValue(rule, key, value);
-
-    if (i == 1) {
+    if (i >= HorusConst::MAX_RULE_SIZE_FOR_HORUS)
+      break;
+    if(!fromRuleToHorusKeyValue(rule, key, value))
+      return;
+    if (i == 0) {
       if (key.setFields == 0) {
         break;
       }
-
       set_fields = key.setFields;
     }
-
     if (key.setFields != set_fields) {
       break;
     }
-
     horus.insert(std::pair<struct HorusRule, struct HorusValue>(key, value));
+    i++;
   }
 }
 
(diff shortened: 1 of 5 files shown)


src/services/pcn-iptables/src/Utils.cpp
```diff
@@ -608,7 +612,9 @@ void Chain::horusFromRulesToMap(
 
   for (auto const &rule : rules) {
     i++;
-    fromRuleToHorusKeyValue(rule, key, value);
+    if (i > HorusConst::MAX_RULE_SIZE_FOR_HORUS)
```
mauriciovasquezbernal: Shouldn't this be `>=`?

src/services/pcn-iptables/src/Utils.cpp
```diff
@@ -608,7 +612,9 @@ void Chain::horusFromRulesToMap(
...
+      break;
+    if(fromRuleToHorusKeyValue(rule, key, value)) {
```
mauriciovasquezbernal: What about refactoring a little bit this?

```
if (!fromRuleToHorusKeyValue(rule, key,value) 
  break;
```

That else and return at the end is quite difficult to read.

src/services/pcn-iptables/src/defines.h
```diff
@@ -124,7 +124,7 @@ const uint8_t DSTPORT = 4;
 // apply Horus mitigator optimization if there are at least # rules
 // matching the pattern, at ruleset begin
 const uint8_t MIN_RULE_SIZE_FOR_HORUS = 1;
-const uint8_t MAX_RULE_SIZE_FOR_HORUS = -1;  // not used
+const uint8_t MAX_RULE_SIZE_FOR_HORUS = 2048;  // not used
```
mauriciovasquezbernal: Is this still not used?

src/services/pcn-iptables/src/datapaths/Iptables_Horus_dp.c
```diff
@@ -133,18 +132,18 @@ static int handle_rx(struct CTXTYPE *ctx, struct pkt_metadata *md) {
...
-  pcn_log(ctx, LOG_DEBUG, "HORUS key: %x. pkt: %x", key.srcIp, pkt->srcIp);
+  pcn_log(ctx, LOG_DEBUG, "Horus srcIp: %I dstIp: %I", pkt->srcIp, pkt->dstIp);
   if (value == NULL) {
     // Miss, goto pipleline
     goto PIPELINE;
   } else {
     // Independently from the final action (ACCEPT or DROP)
     // I have to update the counters
-    if (value->ruleID <= 1024) {
+    if (value->ruleID <= _MAX_RULE_SIZE_FOR_HORUS) {
```
mauriciovasquezbernal: I think here the test has to be `<=`.
The range of ruleID is `[0, _MAX_RULE_SIZE_FOR_HORUS - 1]`

src/services/pcn-iptables/src/Utils.cpp
mbertrone: It should be `>`, but since I reworked the cycle to avoid confusion `>=` works now.

src/services/pcn-iptables/src/Utils.cpp
mbertrone: Thanks for the suggestion. I agree it is more clear.

src/services/pcn-iptables/src/defines.h
mbertrone: Forgot to remove this comment.

src/services/pcn-iptables/src/datapaths/Iptables_Horus_dp.c
mbertrone: The test has to be `<`

mbertrone:
@mauriciovasquezbernal I force-pushed an updated version according to your comments.
Thanks for reviewing it.

src/services/pcn-iptables/src/datapaths/Iptables_Horus_dp.c
mauriciovasquezbernal: I meant `<`, this is very easy to get confused :D

src/services/pcn-iptables/src/defines.h
mauriciovasquezbernal: I just got a compilation warning here, 2048 is too big for a `uint8_t`, I really don't know how this is working.

frisso: @sebymiano Isn't something you already fixed on some of your branch?

sebymiano: I just changed it to `uint32_t` to remove the compilation warning but I haven't tested it.
I guess it should be ok to change.

mbertrone: @mauriciovasquezbernal @frisso @sebymiano It was my mistake during code alignment (the code is correct in the firewall, not in iptables).
I've opened a PR for that #163

PR #152, opened 2019-06-11
SYN Flood Detector

mbertrone:
This PR adds a TCP SYN Flood Detector service, in brief `synflood`.
The purpose of this module is collecting info over the system, that can be used by a remote backend server to detect attacks and anomalies.

Exposed metrics:
-`tcpAttemptFails`: measures the number of TCP connections from state SYN-RCVD to CLOSED or LISTEN.
-`tcpOutRsts`: the number of TCP segments sent, containing RST flag.
-`deliverRatio`: the ratio between the number of IP pkts delivered to application protocols and the total number of received pkts.
-`responseRatio`: the ratio between the number of IP pkts requests to send by application protocols, and the total number of received pkts.


goldenrye:
Matteo,

I have some high-level comments about this feature:
1. does this service also monitor the traffic as intermediate device?
2. does we take the host syn-cookie mechanism into consideration? In this case the state SYN-RECV may not establish

frisso:
Jianwen,
1) No, it operates only if the host is the destination of the attack. It will not operate as "man in the middle", inspecting network traffic that terminates to a third host (at least in the current version).
2) Yes; Matteo can give you some more details.
@mbertrone We should share the design principles, maybe writing some docs, so that we are all aligned here. Also item (1) has to be properly documented.

mbertrone:
A documentation section was added.

frisso:
@acloudiator Ready to be merged from my side.

goldenrye:
Reviewed the code, looks good to me. But I am a little worry about the way the service being used, it is a system-wise service but we have to attach it to a particular interface, I think maybe we should introduce a host type service along with transparent and regular service.

frisso:
Nice idea. But don't delay features, if possible. We ca go with this version now, and create a v2 flavor that works per-interface, which requires polycube internal changes.

acloudiator:
> Nice idea. But don't delay features, if possible. We ca go with this version now, and create a v2 flavor that works per-interface, which requires polycube internal changes.

Added to Roadmap #153

PR #154, opened 2019-06-17
pcn-synflood updated

francescomessina:
- removed the dependency with nstat and awk
 - conversion of parameters from uint64 to strings
   (useful for running the service within namespaces and being able to represent float numbers)


src/services/pcn-synflood/swagger_api/synflood_swagger
```diff
@@ -714,9 +712,8 @@
                         "description": "OK: Successful operation",
                         "schema": {
                             "readOnly": true,
-                            "type": "integer",
-                            "description": "pkts delivered to application over total received pkts",
-                            "format": "uint64"
+                            "type": "string",
+                            "description": "pkts delivered to application over total received pkts"
```
frisso: Ratio between the number of IP pkts delivered to application protocols and the total number of received pkts

src/services/pcn-synflood/swagger_api/synflood_swagger
```diff
@@ -759,9 +756,8 @@
                         "description": "OK: Successful operation",
                         "schema": {
                             "readOnly": true,
-                            "type": "integer",
-                            "description": "pkts requests to send over total received pkts",
-                            "format": "uint64"
+                            "type": "string",
+                            "description": "pkts requests to send over total received pkts"
```
frisso: Ratio between the number of IP pkts requests to send by application protocols and the total number of received pkts

src/services/pcn-synflood/src/Stats.cpp
```diff
@@ -12,67 +12,115 @@
...
 
 Stats::~Stats() {}
 
-const uint64_t ratioMultiplier = 1000;
-
-uint64_t Stats::getTcpattemptfails() {
+std::string Stats::getTcpattemptfails() {
   try {
-    return std::stoull(Synflood::exec("nstat -a | grep TcpAttemptFails | awk '{print $2}'"));
+    return getValue(7, 7);
```
frisso: It would be better to use some #define here, to make the code more clear.

frisso:
LGTM. Thanks!

src/services/pcn-synflood/src/base/StatsBase.cpp
```diff
@@ -25,10 +25,24 @@ void StatsBase::update(const StatsJsonObject &conf) {
 StatsJsonObject StatsBase::toJsonObject() {
```
mauriciovasquezbernal: Hello @francescomessina, It is better to override this function in src/Stats.cpp. Generally speaking files under base should not be changed by the user.

francescomessina: Hello @mauriciovasquezbernal, also in this particular case?  that we want to change the behavior of the container always

francescomessina: Done. Thanks @mauriciovasquezbernal

frisso: @mauriciovasquezbernal Is it ok now?

mauriciovasquezbernal: Yes, it is.

frisso: Can you approve? :-)

PR #155, opened 2019-06-18
Fix issues when loading full topology with a single command

mauriciovasquezbernal:
https://github.com/polycube-network/polycube/pull/126 introduced the possibility of loading a list of cubes present in a yaml or json file. Although that PR worked in the simple cases, it was not possible to load complex topologies.  This PR addresses many of the problems present when creating a full topology from the configuration file:

- set the ports' peer after all the cubes have been created
- add a list of transparent cubes as property in ports, so on creation time this is possible to provide the list of cubes that should be attached to a given port
- fix a bug on the router that created the routing table before the ports
- extend the nat so it is able to reconfigure the masquerade rules when this is attached to a port

This PR is also useful for https://github.com/polycube-network/polycube/pull/141 as all the bugs present solved here affects that PR.


mauriciovasquezbernal:
We decided this is quite related to https://github.com/polycube-network/polycube/pull/141 so we are moving the contents of this PR there.

PR #158, opened 2019-06-25
 Alignment between port/transparent_cube and netdev (when they are connected)

francescomessina:
This PR adds:

A. The methods to align a port or a transparent cube with a netdev, when they are connected to each other.
The methods added are:
1) `subscribe_parameter(caller, parameter, callback function)` = subscribe the port or the transparent cube to the change of the parameter. If the netdev changes that parameter it is called the callback function.
2) `unsubscribe_parameter(caller, parameter)` = cancel the subscription of the port or the transparent cube by changing the parameter.
3) `set_parameter(parameter, value)` = update the parameter on the netdev with the new value.

B. The methods to align the parameters between a transparent cube and a port, when the cube is attached to the port.
The methods in this case are:
4) `subscribe_parent_parameter(parameter, callback function)`.
5) `unsubscribe_parent_parameter(parameter)`.

C. The router ports are extended because when a port is connected to a netdev, the parameters between port and netdev automatically align (same MAC, same IP and same Netmask)

### DIFF
--- a/src/libs/polycube/include/polycube/services/cube_iface.h
+++ b/src/libs/polycube/include/polycube/services/cube_iface.h
@@ -20,6 +20,7 @@
 #include "polycube/services/guid.h"
 #include "polycube/services/port_iface.h"
 #include "polycube/services/table.h"
+#include "polycube/services/types.h"
 
 #include <map>
 #include <string>
@@ -95,7 +96,6 @@ class CubeIface : virtual public BaseCubeIface {
 class TransparentCubeIface : virtual public BaseCubeIface {
  public:
   virtual void set_next(uint16_t next, ProgramType type) = 0;
-  virtual std::string get_parent_parameter(const std::string &parameter) = 0;
   virtual void set_parameter(const std::string &parameter,
                              const std::string &value) = 0;
   virtual void send_packet_out(const std::vector<uint8_t> &packet, Sense sense,
@@ -103,6 +103,11 @@ class TransparentCubeIface : virtual public BaseCubeIface {
 
   virtual void set_conf(const nlohmann::json &conf) = 0;
   virtual nlohmann::json to_json() const = 0;
+
+  virtual void subscribe_parent_parameter(const std::string &param_name,
+                                        ParameterEventCallback &callback) = 0;
+  virtual void unsubscribe_parent_parameter(const std::string &param_name) = 0;
+  virtual std::string get_parent_parameter(const std::string &parameter) = 0;
 };
 }
 }
--- a/src/polycubed/src/extiface.cpp
+++ b/src/polycubed/src/extiface.cpp
@@ -23,9 +23,13 @@
 #include "utils/netlink.h"
 
 #include <iostream>
+#include <net/if.h>
 
 namespace polycube {
 namespace polycubed {
+
+const std::string PARAMETER_MAC = "MAC";
+const std::string PARAMETER_IP = "IP";
 
 std::set<std::string> ExtIface::used_ifaces;
 
@@ -39,6 +43,9 @@ ExtIface::ExtIface(const std::string &iface)
   }
 
   used_ifaces.insert(iface);
+
+  // Save the ifindex
+  ifindex_iface = if_nametoindex(iface.c_str());
 }
 
 ExtIface::~ExtIface() {
@@ -147,10 +154,6 @@ PeerIface *ExtIface::get_peer_iface() {
   return peer_;
 }
 
-std::string ExtIface::get_parameter(const std::string &parameter) {
-  throw std::runtime_error("get parameter not implemented in netdev");
-}
-
 void ExtIface::set_next_index(uint16_t index) {
   std::lock_guard<std::mutex> guard(iface_mutex_);
   set_next(index, ProgramType::INGRESS);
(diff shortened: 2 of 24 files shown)


francescomessina:
@frisso the answers are:

1) There is no need to scan all the ports, because all the ports receive the notification and check if the notification concerns their peer (after checking that the peer they are connected to is a netdev), if so they react.

2) works with an arbitrary number of ports connected to netdev, in the specific case of the router the subscribe occurs when the port is created and reacts to notifications only if the port is connected to a netdev

Anyway I'm adding comments to clarify everything, and I'm editing the code as agreed with Mauricio

src/services/pcn-router/src/Router.cpp
```diff
@@ -775,6 +775,75 @@ void Router::addPorts(const std::string &name, const PortsJsonObject &conf) {
   add_port<PortsJsonObject>(name, conf);
   if (get_shadow())
     getPorts(name)->setMac(getPorts(name)->getMac());
+
+  // Creates function objects containing the callback functions
+  // to be called when there are notifications on one netdev
+  std::function<void(const std::string, const std::string)> f_ip = std::bind(&Router::iface_ip_notification,
```
mauriciovasquezbernal: I think you can use a lambda function here. They are good in these cases because this avoid to create a new method that will make the code difficult to read.

src/services/pcn-router/src/Router.cpp
```diff
@@ -775,6 +775,75 @@ void Router::addPorts(const std::string &name, const PortsJsonObject &conf) {
...
+  // Register the new port to IP and MAC notifications arriving from netdevs
+  get_port(name)->subscribe_event("IP_ADDRESS", f_ip);
+  get_port(name)->subscribe_event("MAC_ADDRESS", f_mac);
+}
+
+// Callback function for ip notification
+void Router::iface_ip_notification(const std::string name_iface, const std::string info_ipAddress) {
+  std::lock_guard<std::mutex> guard(router_mutex);
+
+  // Check if the interface is connected to this cube
```
mauriciovasquezbernal: Why do you need to check that the interface is connected to that cube?
If you receive a notification of an interface that is not connected to the cube it means there is a bug somewhere. IMO this check is useless.

src/services/pcn-router/src/Router.cpp
```diff
@@ -775,6 +775,75 @@ void Router::addPorts(const std::string &name, const PortsJsonObject &conf) {
...
+
+        char split_char = '/';
+        for (std::string each; std::getline(split, each, split_char); info.push_back(each));
+        std::string new_ip = info[0];
+        std::string new_netmask = utils::get_netmask_from_CIDR(std::stoi(info[1]));
+
+        std::string old_ip = port->getIp();
+        std::string old_netmask = port->getNetmask();
+
+        logger()->info("Align ip and netmask of port {0} with those of interface {1}", port->name(), name_iface);
```
mauriciovasquezbernal: info is not the correct logging level, it's too verbose.

src/services/pcn-router/src/Router.cpp
```diff
@@ -775,6 +775,75 @@ void Router::addPorts(const std::string &name, const PortsJsonObject &conf) {
...
+
+  // Creates function objects containing the callback functions
+  // to be called when there are notifications on one netdev
+  std::function<void(const std::string, const std::string)> f_ip = std::bind(&Router::iface_ip_notification,
+                                this, std::placeholders::_1, std::placeholders::_2);
+  std::function<void(const std::string, const std::string)> f_mac = std::bind(&Router::iface_mac_notification,
+                                this, std::placeholders::_1, std::placeholders::_2);
+
+  // Register the new port to IP and MAC notifications arriving from netdevs
+  get_port(name)->subscribe_event("IP_ADDRESS", f_ip);
```
mauriciovasquezbernal: This looks a little bit weird to me, why to handle all this logic at the Router level?, This is related to ports, so all of this should be moved to the Ports class.

src/libs/polycube/include/polycube/services/cube_iface.h
```diff
@@ -103,6 +103,11 @@ class TransparentCubeIface : virtual public BaseCubeIface {
 
   virtual void set_conf(const nlohmann::json &conf) = 0;
   virtual nlohmann::json to_json() const = 0;
+
+  virtual void subscribe_event(const std::string &param_name,
+          std::function<void(const std::string, const std::string)> callback) = 0;
```
mauriciovasquezbernal: `std::function<void(const std::string, const std::string)> callback` is used in many places, maybe it is a good idea to define a new type for this.

src/polycubed/src/port.cpp
```diff
@@ -287,5 +297,47 @@ void Port::unconnect(PeerIface &p1, PeerIface &p2) {
...
+    subscription_list.emplace(param_name, callback);
+
+    // Check if the port already has a peer
+    ExtIface* extiface = dynamic_cast<ExtIface*>(get_peer_iface());
+    if (extiface)
+      extiface->subscribe_event(param_name, callback);
+
+  } else {
+    logger->error("subscribeEvent: Parameter name {0} not available", param_name);
+    throw std::runtime_error("subscribeEvent: Parameter name " + param_name + " not available");
```
mauriciovasquezbernal: this is bad a practice to log and throw an error, just throw the error and the one calling this function should perform the logging of the exception.

src/polycubed/src/extiface.cpp
```diff
@@ -254,5 +255,119 @@ std::string ExtIface::get_iface_name() const {
...
+      std::istringstream old_split(old_value);
+      std::vector<std::string> old_info;
+      for (std::string each; std::getline(old_split, each, split_char); old_info.push_back(each));
+      std::string old_ip = old_info[0];
+      int old_prefix = std::stoi(old_info[1]);
+      Netlink::getInstance().unset_iface_ip(iface_, old_ip, old_prefix);
+    }
+
+    // set new ip/prefix
+    Netlink::getInstance().set_iface_ip(iface_, ip, prefix);
```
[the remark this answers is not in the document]
francescomessina: It add a new IP

francescomessina:
Hello @mauriciovasquezbernal, thanks for the review.
I followed your instructions and I changed everything as requested, let me know if it's ok :)

src/polycubed/src/extiface.cpp
```diff
@@ -254,5 +255,119 @@ std::string ExtIface::get_iface_name() const {
   return iface_;
 }
 
+/* Interfaces alignment */
+void ExtIface::subscribe_event(const std::string &param_name,
+      std::function<void(const std::string, const std::string)> callback) {
+  if (param_name == "MAC_ADDRESS") {
```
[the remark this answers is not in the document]
francescomessina: I defined the strings and changed the name from "IP_ADDREES" and "MAC_ADDRESS" to "IP" and "MAC".
In case of changes, it is sufficient to modify the definition.

src/polycubed/src/utils/netlink.cpp
```diff
@@ -958,5 +961,58 @@ void Netlink::move_iface_into_ns(const std::string &iface, int fd) {
...
+    }
+  }
+  close(fd);
+
+  uint64_t mac_;
+  memcpy(&mac_, mac, sizeof mac_);
+  return polycube::service::utils::be_uint_to_mac_string(mac_);
+}
+
+std::string Netlink::get_iface_ip(const std::string &iface) {
```
mauriciovasquezbernal: Did you test this function? It throws `IP Address is not in a valid format` in some experiments I am doing.

src/polycubed/src/extiface.cpp
```diff
@@ -254,5 +261,99 @@ std::string ExtIface::get_iface_name() const {
...
+    netlink_notification_new_link = [&](int ifindex, const std::string &iface) {
+      // Check if the iface in the notification is that of the ExtIface
+      if (ExtIface::get_iface_name() == iface) {
+        std::string mac_iface = Netlink::getInstance().get_iface_mac(iface);
+        ExtIface::callback_new_macAddress(iface, mac_iface);
+      }
+    };
+    // subscribe to the netlink event for MAC
+    netlink_index_new_link = Netlink::getInstance().registerObserver(
+          Netlink::Event::LINK_ADDED, netlink_notification_new_link);
```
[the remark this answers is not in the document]
francescomessina: Yes this is right. The netlink message is of this type. Changing a MAC address is seen as removing the old link and adding a new link

francescomessina:
> On the technical side, did you test this solution?, I tried some simple configuration and it didn't work for me.
> 
> ```
> polycubectl router add r0
> polycubectl r0 ports add p1 ip=10.0.0.1 netmask=255.255.0.0 
> polycubectl r0 ports p1 set peer=veth1 <--- gives "IP Address is not in a valid format"
> ```

@mauriciovasquezbernal Yes I tested the solution, and I know that the "IP Address is not in a valid format" message is thrown in this case. 
This happens because your `veth1` does not have an IP address when you try to connect / set peer.
If you try with a netdev with all the parameters the code works.

Anyway, what behavior we give to the code in this case? (router ports require mandatory IP and netmask, that's why I have that message)

src/polycubed/src/extiface.h
```diff
@@ -43,14 +44,24 @@ class ExtIface : public PeerIface {
...
   void set_peer_iface(PeerIface *peer) override;
   PeerIface *get_peer_iface() override;
-  std::string get_parameter(const std::string &parameter) override;
 
   void set_next_index(uint16_t index);
   bool is_used() const;
 
   std::string get_iface_name() const;
 
+  void subscribe_parameter(const std::string &caller, const std::string &param_name, paremeter_event_callback callback) override;
```
mauriciovasquezbernal: Please split these too long lines to be <80 characters long.

src/polycubed/src/port.cpp
```diff
@@ -287,5 +293,51 @@ void Port::unconnect(PeerIface &p1, PeerIface &p2) {
...
+  ExtIface* extiface = dynamic_cast<ExtIface*>(get_peer_iface());
+  if (extiface)
+    extiface->subscribe_parameter(name_, param_name, callback);
+}
+
+void Port::unsubscribe_peer_parameter(const std::string &param_name) {
+  // Remove event from the list
+  subscription_list.erase(param_name);
+
+  // Only ubscribe if peer is ExtIface
```
mauriciovasquezbernal: `ubscribe` -> `unsubscribe`

src/polycubed/src/port.cpp
```diff
@@ -141,8 +137,18 @@ void Port::set_peer(const std::string &peer) {
...
+    std::lock_guard<std::mutex> guard(port_mutex_);
+    peer_ = peer;
+  }
+
+  // Check if peer is a ExtIface
+  ExtIface* extiface = dynamic_cast<ExtIface*>(get_peer_iface());
+  if (extiface) {
+    // Subscribe to the list of events
+    for (auto it : subscription_list)
+      extiface->subscribe_parameter(name_, it.first, it.second);
```
mauriciovasquezbernal: `name_` is being used as caller_id here, this is wrong because this is possible to have a transparent cube and a port with the same name. Maybe this is a good place to use the uuid of those elements, so we don't have to bother about possible collisions.

src/polycubed/src/port.h
```diff
@@ -80,6 +79,17 @@ class Port : public polycube::service::PortIface, public PeerIface {
   static void connect(PeerIface &p1, PeerIface &p2);
   static void unconnect(PeerIface &p1, PeerIface &p2);
 
+  // Implemenation of parameter subscription in PeerIface.
```
mauriciovasquezbernal: `Implemenation` -> `Implementation`

src/polycubed/src/transparent_cube.cpp
```diff
@@ -73,24 +73,18 @@ void TransparentCube::set_next(uint16_t next, ProgramType type) {
 
 void TransparentCube::set_parent(PeerIface *parent) {
   parent_ = parent;
-  if (parent) {
+  if (parent_) {
     attach_();
+    for (auto it : subscription_list) {
```
mauriciovasquezbernal: Use references in these loops where possible to about many copies.

src/polycubed/src/transparent_cube.cpp
```diff
@@ -163,5 +157,40 @@ nlohmann::json TransparentCube::to_json() const {
...
+}
+
+std::string TransparentCube::get_parent_parameter(const std::string &param_name) {
+  if (!parent_) {
+    throw std::runtime_error("cube is not attached");
+  }
+  return parent_->get_parameter(param_name);
+}
+
+//void TransparentCube::set_parent_parameter(const std::string &param_name, const std::string &value) {
```
mauriciovasquezbernal: Please avoid to submit commented code in a PR.

src/services/pcn-router/src/Ports.cpp
```diff
@@ -89,9 +89,69 @@ Ports::Ports(polycube::service::Cube<Ports> &parent,
     PortsSecondaryip::createInControlPlane(*this, addr.getIp(),
                                            addr.getNetmask(), addr);
   }
+
+  // lambda function to align IP and Netmask between Port and ExtIface
+  paremeter_event_callback f_ip;
+  f_ip = [&](const std::string name_iface, const std::string info_ipAddress) {
+    if (Ports::peer() == name_iface) {
```
mauriciovasquezbernal: Any specific reason you are using the `Ports::` prefix, if not please remove it.

src/services/pcn-router/src/Ports.cpp
```diff
@@ -159,10 +219,15 @@ void Ports::setIp(const std::string &value) {
     // set new ip
     if (parent_.get_shadow()) {
       int prefix = get_netmask_length(getNetmask());
-      std::function<void()> doThis = [&]{parent_.netlink_instance_router_.set_iface_ip(getName(), value, prefix);};
+      std::function<void()> doThis = [&]{parent_.netlink_instance_router_.add_iface_ip(getName(), value, prefix);};
```
mauriciovasquezbernal: Shouldn't it be part of the first commit of this PR where these names are changed in netlink. If you introduce this change here it means the code will not compile after the first commit is applied, what is a pity because it will make git bisect difficult to use.

src/services/pcn-router/src/Ports.cpp
```diff
@@ -159,10 +219,15 @@ void Ports::setIp(const std::string &value) {
...
       polycube::polycubed::Namespace namespace_ = polycube::polycubed::Namespace::open("pcn-" + parent_.get_name());
       namespace_.execute(doThis);
     }
+
+    // Align Port with the peer interface
+    int prefix = get_netmask_length(getNetmask());
+    std::string info_address = ip_ + "/" + std::to_string(prefix);
```
mauriciovasquezbernal: info_address is not the best name for this. something like `cidr` could work better.

francescomessina:
Done

frisso:
@mauriciovasquezbernal LGTM. Should we merge?

mauriciovasquezbernal:
> @mauriciovasquezbernal LGTM. Should we merge?

I am still reviewing this, will approve when I think this is ready.

> Btw, Polycubed, when starts, still says something "pre_transparent_services+". Shouldn't we update this?

That 's the version that is automatically taken from the git repo when it is compiled. That's was our last tag at that time, the + indicates that further changes have been done to it. Once we have a new tag it'll be considered as the new version by the compilation system.

frisso:
In my view, this was the last very important feature missing in Polycube.
Thanks, @mauriciovasquezbernal and @francescomessina !

PR #159, opened 2019-06-26
Update AUTHORS

frisso:
Added Riccardo Marchi (added stateful capabilities to polycubed), Gianluca Scopelliti (updated and improved the bridge service) and Nico Caprioli (parsing of YANG datamodels at run-time and dynamic update of the REST web server).


frisso:
> Let's add contributors' names timely in the future.

Good point, definitely.
My fault.

PR #160, opened 2019-06-27
Revert polycubed persistency support

mauriciovasquezbernal:
We found some bugs that make polycube unusable at this point.  So, revert this
until we have a more robust solution.

There is not a clear idea yet of what are the bugs, but many tests started to fail after this was merged:
http://130.192.225.104:8080/blue/organizations/jenkins/polycube-test/detail/master/94/pipeline/


mauriciovasquezbernal:
Almost all tests are passing[1], there are some problems with the shadow router but it should be another issue, my suggestion is to merge this now to make polycube usable again.

[1] http://130.192.225.104:8080/blue/organizations/jenkins/polycube-test/detail/PR-160/1/pipeline/

mauriciovasquezbernal:
It appears @richiMarchi already has a solution for it, waiting for his PR then.

acloudiator:
> There is not a clear idea yet of what are the bugs, but many tests started to fail after this was merged:

@mauriciovasquezbernal @mbertrone  AFAIK, the merge to master for any PR is blocked if the docker build is failed, isn't it? What would be the case here in PR#141?!

mauriciovasquezbernal:
> > There is not a clear idea yet of what are the bugs, but many tests started to fail after this was merged:
> 
> @mauriciovasquezbernal @mbertrone AFAIK, the merge to master for any PR is blocked if the docker build is failed, isn't it? What would be the case here in PR#141?!

The merge is not blocked by the docker build, additional this build does not perform the tests, those are done with jenkins.

So far this is manual procedure, one before merging should check is tests are ok.

acloudiator:
> So far this is manual procedure, one before merging should check is tests are ok.

Exactly you made my point, the merge was done without the test. Not getting into details, but certainly, it's a lesson learned @polycube-network/pcn-maintainers 

Also, we might want to think of automating this procedure, rather than manual.

mauriciovasquezbernal:
The fix for the problem appears not to be that trivial, so I suggest to merge this PR in order to fix polycube, once the fix is ready we can reintroduce the whole feature.

acloudiator:
> The fix for the problem appears not to be that trivial, so I suggest to merge this PR in order to fix polycube, once the fix is ready we can reintroduce the whole feature.

@frisso If the change isn’t that straightforward then I agree with @mauriciovasquezbernal .
Would you please approve and merge, if you too agree on this.

PR #163, opened 2019-06-27
pcn-iptables: fix wrong const type

mbertrone:
wrong const type was committed by mistake during porting of HORUS functionalities.
This PR fixes this issue.

### DIFF
--- a/src/services/pcn-iptables/src/defines.h
+++ b/src/services/pcn-iptables/src/defines.h
@@ -123,8 +123,8 @@ const uint8_t DSTPORT = 4;
 
 // apply Horus mitigator optimization if there are at least # rules
 // matching the pattern, at ruleset begin
-const uint8_t MIN_RULE_SIZE_FOR_HORUS = 1;
-const uint8_t MAX_RULE_SIZE_FOR_HORUS = 2048;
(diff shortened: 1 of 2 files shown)


acloudiator:
Labeled as `under-review` so that we can test and verify it before merge to master.

src/services/pcn-iptables/src/defines.h
```diff
@@ -124,7 +124,7 @@ const uint8_t DSTPORT = 4;
 // apply Horus mitigator optimization if there are at least # rules
 // matching the pattern, at ruleset begin
 const uint8_t MIN_RULE_SIZE_FOR_HORUS = 1;
-const uint8_t MAX_RULE_SIZE_FOR_HORUS = 2048;
+const uint32_t MAX_RULE_SIZE_FOR_HORUS = 2048;
```
frisso: @mbertrone Looks strange to me that the previous line is still uint8_t. If nothing is affected, I would propose to move it to uint32_t as well, as this looks more coherent and it doesn't raise questions.

mbertrone: Agree, having `uint32_t` is more coherent.
I've force-pushed this change.

PR #164, opened 2019-06-27
fix save_load_topology

richiMarchi:
save_load_topology had a bug while inserting two arrays in a single command. This fixes it


mauriciovasquezbernal:
There are still many tests failing: http://130.192.225.104:8080/blue/organizations/jenkins/polycube-test/detail/PR-164/1/pipeline, I recommend you to check which ones are failing and try them locally.

PR #166, opened 2019-06-28
load old topology at startup and save each modification to file

richiMarchi:
This feature allows to load a topology from file at startup, either from the default folder or from a user selected file (ex. polycubed --cubes-dump ~/Desktop/cubes.yaml), keeping the whole configuration in memory.
After each modification in any of the topology of the cube, the configuration in memory is updated and another thread saves it to file updating it too.
It is possible to start the daemon with an empty topology by using the flag --cubes-init: if the file contains a topology, at the first update it is overwritten.
It is possible to start the daemon without the dumping to file functionality by using the flag --cubes-nodump.

### DIFF
--- a/src/polycubed/src/rest_server.cpp
+++ b/src/polycubed/src/rest_server.cpp
@@ -17,6 +17,7 @@
 
 #include <string>
 #include <vector>
+#include <fstream>
 
 #include "polycubed_core.h"
 #include "service_controller.h"
@@ -26,12 +27,16 @@
 #include "polycube/services/response.h"
 #include "server/Resources/Data/AbstractFactory.h"
 #include "server/Server/ResponseGenerator.h"
+#include "config.h"
+#include "cubes_dump.h"
 
 namespace polycube {
 namespace polycubed {
 
 std::string RestServer::whitelist_cert_path;
 std::string RestServer::blacklist_cert_path;
+
+const std::string RestServer::base = "/polycube/v1/";
 
 // start http server for Management APIs
 // Incapsultate a core object // TODO probably there are best ways...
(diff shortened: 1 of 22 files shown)


acloudiator:
I don’t think our docker-build test covers this PR, is it?
Has to be tested and validated before merging to master.

goldenrye:
Hi Riccardo,

I create a firewall fw0 followed by attach the firewall to the one of interface: veth1, only firewall is dumped to the cubes.yaml, the topology is not dumped.

jpi@polycube-dev-2:~/polycube/build$ polycubectl show cubes

firewall:
 name  uuid                                  service-name  type  loglevel  parent
 fw0   c12ec9d7-4063-4a43-a8b3-36361239d127  firewall      TC    INFO      veth1

jpi@polycube-dev-2:~/polycube/build$ polycubectl show topology
loglevel  name  parent  service-name  type  uuid
INFO      fw0   veth1   firewall      TC    c12ec9d7-4063-4a43-a8b3-36361239d127

jpi@polycube-dev-2:~/polycube/build$ cat cubes.yaml
[
  {
    "service-name": "firewall",
    "name": "fw0",
    "type": "TC",
    "loglevel": "INFO",
    "interactive": true,
    "chain": [
      {
        "default": "forward",
        "name": "ingress"
      },
      {
        "default": "forward",
        "name": "egress"
      }
    ]
  }
]

richiMarchi:
Hi @goldenrye ,
it has been decided that netdevs are not dumped in the topology at the moment.

src/polycubed/src/rest_server.cpp
```diff
@@ -152,6 +157,59 @@ void RestServer::shutdown() {
...
+  for (auto &[peer1, peer2] : peers) {
+    core.try_to_set_peer(peer1, peer2);
+  }
+}
+
+void RestServer::load_last_topology() {
+  logger->info("loading cubes from {}", configuration::config.getCubesDumpFilePath());
+  std::ifstream topologyFile(configuration::config.getCubesDumpFilePath());
+  if (!topologyFile.is_open()) {
+    logger->warn("error opening dump file: {}", configuration::config.getCubesDumpFilePath());
```
mauriciovasquezbernal: I think this message can be deleted, if the file is not present it means there is nothing to be loaded.

frisso:
> Hi @goldenrye ,
> it has been decided that netdevs are not dumped in the topology at the moment.

In fact, the fact is not that we have "decided", but that there are currently technical issues for this. So, we "decided" not to add this feature right now. Due to this (and other small problems) we would suggest to keep this feature ("stateful polycubed") as optional, so not turned on by default.

acloudiator:
> > Hi @goldenrye ,
> > it has been decided that netdevs are not dumped in the topology at the moment.
> 
> In fact, the fact is not that we have "decided", but that there are currently technical issues for this. So, we "decided" not to add this feature right now. Due to this (and other small problems) we would suggest to keep this feature ("stateful polycubed") as optional, so not turned on by default.

Hi @frisso 
Can we list the technical and other issues related to this feature? 

AFAIK, we collectively agreed to have this feature as `a must` as we need to maintain the state. However, the challenges would always be there and my recommendation is to list them down, clearly mention as known issues and add to our roadmap and plan to fix it.

Please share your thoughts. Thanks!

frisso:
> Please share your thoughts. Thanks!

Agreed with you. Limitations have been documented in the 'stateful' section of the .rst file.

mauriciovasquezbernal:
@frisso it is better to perform a git rebase instead of merging master here, that helps a little bit to keep things more organized.

frisso:
> @frisso it is better to perform a git rebase instead of merging master here, that helps a little bit to keep things more organized.

Fine for me. but not having time know to do personally, I have to go to sleep for tomorrow's trip, in a few (really few) hours.

acloudiator:
> I tested and it appears to work good.

@mauriciovasquezbernal Was it a manual test or tested using Jenkins? I would like to follow the same instruction in our lab.

PR #167, opened 2019-07-05
pcn-k8s: use go modules to manage dependencies

mauriciovasquezbernal:
So far the dependencies for the pcn-k8s agent have been managed by saving all
of them in the vendor folder manually without using any tool like go dep.
This approach makes it very hard to update and manage them.

Go mod is a quite new mechanishm to manage golang dependencies, it will become
the default in the future, for this reason this commit uses it instead of
using a mid-term solution like go dep.

This new approach allows to remove the vendor folder from the repo, it makes
it a lot lighter as ~100MB and ~3M lines of code are removed. Also the building
system is simplified and some build workarounds are removed.

This commit uses the following version of the main dependencies:
- github.com/aws/aws-sdk-go -> v1.20.15
- github.com/containernetworking/cni -> v0.7.1
- k8s.io/client-go -> kubernetes-1.15.0
- k8s.io/api -> kubernetes-1.15.0
- k8s.io/utils -> kubernetes-1.15.0
- k8s.io/apimachinery -> kubernetes-1.15.0


mauriciovasquezbernal:
@SunSince90 could you please test if this work with https://github.com/polycube-network/polycube/pull/157?

PR #169, opened 2019-07-05
polycubed: fix connect after attach to netdev

mauriciovasquezbernal:
All the logic to allow to connect a cube's port and attach a transparent cube
to a netdev is inplace, before this commit it was possible to connect a port
to a netdev and then attach a transparent cube to the netdev, however the
inverse procedure, first attach and then connect was not working.

case working before:
- polycubectl router add r0
- polycubectl r0 ports add p1 ip=192.168.1.1 netmask=255.255.255.0
- polycubectl connect r0:p1 veth1
- polycubectl nat add nat0
- polycubectl attach nat0 veth1

case fixed by this commit:
- polycubectl nat add nat0
- polycubectl attach nat0 veth1
- polycubectl router add r0
- polycubectl r0 ports add p1 ip=192.168.1.1 netmask=255.255.255.0
- polycubectl connect r0:p1 veth1

### DIFF
--- a/src/polycubed/src/service_controller.cpp
+++ b/src/polycubed/src/service_controller.cpp
@@ -220,22 +220,30 @@ void ServiceController::set_port_peer(Port &p, const std::string &peer_name) {
 
   std::lock_guard<std::mutex> guard(service_ctrl_mutex_);
 
-  std::unique_ptr<ExtIface> iface;
+  std::shared_ptr<ExtIface> iface;
   std::string cube_name, port_name;  // used if peer is cube:port syntax
 
   if (Netlink::getInstance().get_available_ifaces().count(peer_name) != 0) {
-    switch (p.get_type()) {
-    case PortType::TC:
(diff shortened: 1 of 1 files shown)


mauriciovasquezbernal:
Hello Jian Wen, I understand your concern, but have you tested it?

An entry in `ports_to_ifaces` is present when there is a port connected OR a transparent cube attached to a given netdev, if there is a port connected and transparent cubes attached there is just a single entry for that netdev in the list.

(1) when when a transparent cube is attached is has to be tested if a there was already a port connected to it, and viceversa, (2) when a port is connected it has to be tested if there is already a cube attached to it.  Case 1 was implemented and case 2 was missing, that's what this commit fixes.

The code in this commit checks if there is an entry for the given netdev (if there is something attached or connected), and the it checks if there is something connected, if yes it has to fail because only a single port is allowed to be connected to a given netdev.

goldenrye:
Thanks Mauricio, for explaining. I test it works.

PR #176, opened 2019-07-12
cleanup install

mauriciovasquezbernal:
The main purpose of this PR is to modify the install script to avoiding the installation process to be blocked because the user is prompted for an option. This is causing issues in jenkins because tests are blocked.

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -61,22 +61,12 @@ DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 mkdir -p $WORKDIR
 
 $SUDO apt update
-$SUDO apt install -y wget gnupg2
+$SUDO apt install -y wget gnupg2 software-properties-common
 
-echo "Install golang v1.12"
-# TODO: v1.12 still not available in repos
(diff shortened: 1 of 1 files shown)


scripts/install.sh
```diff
@@ -61,22 +61,12 @@ DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
...
-
-$SUDO apt-get install -y software-properties-common
-$SUDO add-apt-repository ppa:longsleep/golang-backports
-$SUDO apt-get update
-$SUDO apt install -y golang-go
-#$SUDO ln -s /usr/bin
+$SUDO apt install -y wget gnupg2 software-properties-common
 
+# golang v1.12 still not available in repos
+$SUDO add-apt-repository ppa:longsleep/golang-backports -y
```
frisso: Mauricio,  L75 ($SUDO apt-get update) is missing here.
Shouldn't we update the repo before installing?

mauriciovasquezbernal: It is now at line 73. (A single update after new sources are added should be enough)

frisso: Ops... it didn't show up in the diff :-(

PR #178, opened 2019-07-16
Update router to the ip/prefix notation, removing the netmask

francescomessina:
This PR updates the router code:
- Changed IP addresses and routes from the notation with the netmask to the notation with prefix (CIDR).
- The new ports do not require a mandatory IP when they are created.
- Changed the name arp-entry in arp-table.
- Adds new methods to the utils library.


src/libs/polycube/src/utils.cpp
```diff
@@ -307,6 +311,49 @@ std::string get_netmask_from_CIDR(const int cidr) {
...
+  if ((ipAddress & mask) == (net & mask))
+    return true;
+  else
+    return false;
+}
+
+std::string get_network_from_ip(const std::string &ip,
+                                const std::string &netmask) {
+  // get the network from ip
+  uint32_t address = ip_string_to_be_uint(ip);
```
frisso: This comment is not related to this PR; however, I feel the name of this function is misleading. In fact, we're not transforming a string into a "big endian" number, but in "network byte order" number.
So, I would kindly ask you to open a new PR and rename the following functions (and add a comment in the code, so that people know that "nbo" is "network byte order"):
- ip_string_to_be_uint ==> ip_string_to_nbo_uint
- be_uint_to_ip_string ==> nbo_uint_to_ip_string
The same for the other functions that operate on MAC addresses.

src/libs/polycube/src/utils.cpp
```diff
@@ -307,6 +311,49 @@ std::string get_netmask_from_CIDR(const int cidr) {
...
+  uint32_t mask = ip_string_to_be_uint(netmask);
+  uint32_t net = address & mask;
+  char buffer[100];
+  sprintf(buffer, "%d.%d.%d.%d", (net) & 0xFF, (net >> 8) & 0xFF,
+          (net >> 16) & 0xFF, (net >> 24) & 0xFF);
+  std::string network(buffer);
+  return network;
+}
+
+std::string from_int_to_hex(int t) {
```
frisso: Is the 'from' in the function name needed? It looks to me that the other functions in utils.cpp do not have 'from'.

src/polycubectl/cliargs/cliargs.go
```diff
@@ -304,22 +305,22 @@ func ParseCLIArgs(args []string) (*CLIArgs, error) {
 }
 
 func (cli *CLIArgs) buildURL() string {
-	var url string
+	var url_str string
 
 	for _, str := range cli.PathArgs {
-		url += str + "/"
+		url_str += url.PathEscape(str) + "/"
```
frisso: Add a comment why the above changes are needed. This applies also to line 316.

src/polycubectl/completion/polycubectl.sh
```diff
@@ -25,6 +25,26 @@ parse_yaml() {
 
 BASE_URL_="http://localhost:9000/polycube/v1/"
 
+# https://stackoverflow.com/a/10660730
```
frisso: Add a comment: what is this function about?

src/polycubed/src/utils/utils.cpp
```diff
@@ -92,6 +92,107 @@ std::map<std::string, json> strip_port_tcubes(json &jcube) {
...
+    v = c - 'A' + 10;
+    return true;
+  } else if ('a' <= c && c <= 'f') {
+    v = c - 'a' + 10;
+    return true;
+  }
+  return false;
+}
+
+inline bool from_hex_to_i(const std::string &s, int i, int cnt, int &val) {
```
frisso: what is "i" in the function name about? int?
I presume the result is in host byte order; in  this case, let's write hbo_int (and add a comment before the function)

src/polycubed/src/utils/utils.cpp
```diff
@@ -92,6 +92,107 @@ std::map<std::string, json> strip_port_tcubes(json &jcube) {
...
+    if (is_hex(s[i], v)) {
+      val = val * 16 + v;
+    } else {
+      return false;
+    }
+  }
+  return true;
+}
+
+std::string decode_url(const std::string &s) {
```
frisso: What this "decode_url" does? Write a comment.

src/services/pcn-router/datamodel/router.yang
```diff
@@ -19,35 +19,19 @@ module router {
...
-        description "Netmask of the port";
-        polycube-base:cli-example "255.255.255.0";
+      leaf ip {
+        type inet:ipv4-prefix;
+        description "IP address and prefix of the port";
+        polycube-base:cli-example "10.0.0.1/24";
       }
 
       list secondaryip {
         description "Secondary IP address for the port";
```
frisso: description "Secondary IP address for the port";
==>
description "Additional IP addresses for the port";

src/services/pcn-router/datamodel/router.yang
```diff
@@ -19,35 +19,19 @@ module router {
...
-          polycube-base:cli-example "207.46.131.1";
-        }
-
-        leaf netmask {
-          type inet:ipv4-address;
-          mandatory true;
-          description "Secondary netmask of the port";
-          polycube-base:cli-example "255.255.255.0";
+          type inet:ipv4-prefix;
+          description "Seconadary IP address and prefix of the port";
```
frisso: description "Seconadary IP address and prefix of the port";
==>
description "Secondary IP address and prefix of the port";

src/services/pcn-router/src/Ports.h
```diff
@@ -16,88 +16,67 @@
...
-
-#include <spdlog/spdlog.h>
 #include <set>
 
 #include "PortsSecondaryip.h"
 
 class Router;
 
-#define SECONDARY_ADDRESS 5
+#define MAX_SECONDARY_ADDRESSES 5
```
frisso: This constant is defined twice.
Either we defined only in one place, or we add a comment (in both places) telling the programmer that this constant has to be updated in both places.

src/services/pcn-router/src/PortsSecondaryip.cpp
```diff
@@ -1,118 +1,54 @@
...
 
 void PortsSecondaryip::createInControlPlane(
-    Ports &parent, const std::string &ip, const std::string &netmask,
+    Ports &parent, const std::string &ip,
     const PortsSecondaryipJsonObject &conf) {
   // This method creates the actual PortsSecondaryip object given thee key
   // param.
 
-  // TODO: a port cannot have more than SECONDARY_ADDRESS secondary addresses.
+  // TODO: a port cannot have more than MAX_SECONDARY_ADDRESSES secondary addresses.
```
frisso: This is important.
If would be nice to add a check in the code so that we cannot set more than MAX_SECONDARY_ADDRESSES.

src/services/pcn-router/src/Router.h
```diff
@@ -54,93 +49,82 @@ struct rt_v {
...
 
-  void remove_route(const std::string &network, const std::string &netmask,
-                    const std::string &nexthop);
+  void remove_route(const std::string &network, const std::string &nexthop);
 
   void remove_all_routes();
 
-  /* Netlink */
+  // Shadow
   std::mutex router_mutex;
```
frisso: Can we be more specific with respect to "router_mutex"?
I cannot understand what this mutex protect about.

src/services/pcn-router/test/bridge/test2_2_4.sh
```diff
@@ -44,7 +44,7 @@ polycubectl router add r2 type=$TYPE
 polycubectl simplebridge add br1 type=$TYPE
 polycubectl simplebridge add br2 type=$TYPE
 
-router_add_port r1 to_br1 10.10.2.1 255.255.255.0
+router_add_port r1 to_br1 10.10.2.1 24
```
frisso: Shouldn't be:
   router_add_port r1 to_br1 10.10.2.1/24
Without the "slash", the new syntax looks definitely not intuitive.
Btw, there was a bug in the CLI that prevented to handle the "/" in the proper way, but Mauricio fixed it some days ago.

src/polycubed/src/utils/utils.cpp
francescomessina: @mauriciovasquezbernal this code is already in the master (you wrote it to make url encoding).
Maybe you know better than me what it does.
PS: I have done the rebase with the master so these files will no longer be modified by me in this PR

src/polycubed/src/utils/utils.cpp
francescomessina: @mauriciovasquezbernal the same here

src/polycubectl/completion/polycubectl.sh
francescomessina: @mauriciovasquezbernal

src/polycubectl/cliargs/cliargs.go
francescomessina: @mauriciovasquezbernal

src/polycubed/src/utils/utils.cpp
frisso: Thanks!

src/services/pcn-router/src/PortsSecondaryip.cpp
```diff
@@ -1,118 +1,55 @@
...
- * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
- * See the License for the specific language governing permissions and
- * limitations under the License.
- */
+/**
+* router API generated from router.yang
+*
+* NOTE: This file is auto generated by polycube-codegen
+* https://github.com/polycube-network/polycube-codegen
+*/
```
goldenrye: Is this file entirely generated by swagger without any modification? If we made some changes, it is better to add some comments here.

src/services/pcn-router/src/PortsSecondaryip.h
```diff
@@ -1,66 +1,41 @@
...
- * See the License for the specific language governing permissions and
- * limitations under the License.
- */
+/**
+* router API generated from router.yang
+*
+* NOTE: This file is auto generated by polycube-codegen
+* https://github.com/polycube-network/polycube-codegen
+*/
+
```
goldenrye: same comment as previous one, if the file is modified by us it is better to add some comment here.

frisso:
> 1. some files are auto-generated by swagger and I see the header in a few files was replace by "auto-generated by swagger ...", it is good to replace it if the entire file was auto-generated w/o modification; otherwise we should write some comments to indicates it is a modified version;

Good point.
I don't know how to do that, but definitely it's a good suggestion.

> 2. it is better to split this commit to a few small commit, especially for the commit to replace the arp_entry with arp_table, which will make the reviewers to focus on the logical change.

This is just a cosmetic change: we simply renamed the structure in the YANG. Hence, we had to update also the internal structures.

Overall, this looks a very minor PR: mainly cosmetics, i.e., changing the name of the arp table (arp_table, in fact, instead of the ugly name "arp_entry"), and allowing IP addresses to be written as a.b.c.d/prefixlength, which looks simpler than the "old" syntax (ip netmask).
So, in the end, no complex things are changed.

goldenrye:
> it is better to split this commit to a few small commit, especially for the commit to replace the arp_entry with arp_table, which will make the reviewers to focus on the logical change.
> This is just a cosmetic change: we simply renamed the structure in the YANG. Hence, we had to update also the internal structures.
> 
> Overall, this looks a very minor PR: mainly cosmetics, i.e., changing the name of the arp table (arp_table, in fact, instead of the ugly name "arp_entry"), and allowing IP addresses to be written as a.b.c.d/prefixlength, which looks simpler than the "old" syntax (ip netmask).
> So, in the end, no complex things are changed.
> 
Yes the logical changes in this commit is minor, but the lines of code modified is non-trivial due to the renaming, making review of other change difficult.

goldenrye:
- Jianwen 

> On Jul 17, 2019, at 1:34 PM, Jianwen Pi <jianwpi@gmail.com> wrote:
> 
> 
>> it is better to split this commit to a few small commit, especially for the commit to replace the arp_entry with arp_table, which will make the reviewers to focus on the logical change.
>> This is just a cosmetic change: we simply renamed the structure in the YANG. Hence, we had to update also the internal structures.
>> 
>> Overall, this looks a very minor PR: mainly cosmetics, i.e., changing the name of the arp table (arp_table, in fact, instead of the ugly name "arp_entry"), and allowing IP addresses to be written as a.b.c.d/prefixlength, which looks simpler than the "old" syntax (ip netmask).
>> So, in the end, no complex things are changed.
>> 
> Yes the logical changes in this commit is minor, but the lines of code modified is non-trivial due to the renaming, making review of other change difficult.

For this commit I am fine to keep as a single commit.

goldenrye:
> some files are auto-generated by swagger and I see the header in a few files was replace by "auto-generated by swagger ...", it is good to replace it if the entire file was auto-generated w/o modification; otherwise we should write some comments to indicates it is a modified version;
> Good point.
> I don't know how to do that, but definitely it's a good suggestion
> 

I didn’t expect a fancy way to do that, at least whenever we modify a auto-generated file we can manually modify the header as well if possible

src/services/pcn-router/datamodel/router.yang
```diff
@@ -19,35 +19,19 @@ module router {
...
   uses "polycube-standard-base:standard-base-yang-module" {
      augment ports {
-       leaf ip {
-        type inet:ipv4-address;
-        mandatory true;
-        description "IP address of the port";
-        polycube-base:cli-example "207.46.130.1";
-      }
-
-      leaf netmask {
```
acloudiator: No more CLI support for netmask?

src/services/pcn-router/src/ArpTable.cpp
```diff
@@ -0,0 +1,54 @@
...
+    : ArpTableBase(parent) {
+  logger()->info("Creating ArpTable instance");
+}
+
+ArpTable::ArpTable(Router &parent, const std::string &mac,
+                   const std::string &ip, const std::string &interface)
+    : ArpTableBase(parent), mac_(mac), ip_(ip), interface_(interface) {}
+
+ArpTable::~ArpTable() {}
+
```
acloudiator: Is the JSONObject implementation missing or dropped for some reason?

```	
void ArpEntry::update(const ArpEntryJsonObject &conf) {		
  // This method updates all the object/parameter in ArpEntry object specified		
  // in the conf JsonObject.		
  // You can modify this implementation.		
  setInterface(conf.getInterface());		
  setMac(conf.getMac());		
}		
ArpEntryJsonObject ArpEntry::toJsonObject() {		
  ArpEntryJsonObject conf;		
  conf.setInterface(getInterface());		
  conf.setMac(getMac());		
  conf.setAddress(getAddress());		
  return conf;		ArpTable::~ArpTable() {}
}
```

frisso: This is part of the automatically generated code. So, it should be correct.

src/services/pcn-router/src/PortsSecondaryip.cpp
frisso: Yes, it is.
The header changes because the router used a previous version of the code generation tool; in the latest version, we changed also the autogenerated header.

src/services/pcn-router/src/ArpTable.cpp
francescomessina: In the new version of the generator it is automatically generated.
For this reason many functions including JSONObject are now found in the folder "/pcn-router/src/base" in the related files.

src/libs/polycube/src/utils.cpp
```diff
@@ -296,17 +296,34 @@ uint32_t get_netmask_length(const std::string &netmask_string) {
...
 
+void split_ip_and_prefix(const std::string &ip_and_prefix,
+                        std::string &ip_address, std::string &netmask) {
+  // ip_and_prefix = ip_address/prefix
+  std::istringstream split(ip_and_prefix);
+  std::vector<std::string> info;
+  char split_char = '/';
+  for (std::string each; std::getline(split, each, split_char);
+       info.push_back(each))
+    ;
```
mauriciovasquezbernal: Why is that `;` at this position?

mauriciovasquezbernal: That for loop is very condensed, this is quite difficult to read for me.

PR #180, opened 2019-07-17
fix polycubed.conf default configuration and cubes dump bug

richiMarchi:
Commented default cubes-dump-file configuration to avoid affecting
parameters validation and removed useless variable assignment.
Check if the dump is enabled with the variable set in config.cpp
instead of checking if CubesDump instance is created or not.

### DIFF
--- a/src/polycubed/src/server/Resources/Endpoint/ListResource.cpp
+++ b/src/polycubed/src/server/Resources/Endpoint/ListResource.cpp
@@ -20,6 +20,7 @@
 #include <utility>
 #include <vector>
 
+#include "../../config.h"
 #include "../../Server/ResponseGenerator.h"
 #include "../Body/JsonNodeField.h"
 #include "../Body/ListKey.h"
@@ -182,11 +183,10 @@ void ListResource::CreateReplaceUpdateWhole(
       errors.push_back(resp);
     }
     // check if the operation completed successfully and in case update the configuration
-    if (isOperationSuccessful(resp.error_tag)) {
-      if (auto d = core_->get_cubes_dump()) {
-        d->UpdateCubesConfig(request.resource(), jbody, keys, op,
-                ResourceType::ListResource);
-      }
+    if (isOperationSuccessful(resp.error_tag) &&
+        configuration::config.getCubesDumpEnabled()) {
+      core_->get_cubes_dump()->UpdateCubesConfig(request.resource(), jbody,
+              keys, op, ResourceType::ListResource);
     }
   }
   Server::ResponseGenerator::Generate(std::move(errors), std::move(response));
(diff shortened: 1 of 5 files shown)


src/polycubed/src/server/Resources/Endpoint/ListResource.cpp
```diff
@@ -19,6 +19,7 @@
 #include <string>
 #include <utility>
 #include <vector>
+#include <../../config.h>
```
frisso: I would move this one line down, to keep an empty line between standard headers and the ones of the project.

src/polycubed/src/server/Resources/Endpoint/ListResource.cpp
```diff
@@ -183,8 +184,8 @@ void ListResource::CreateReplaceUpdateWhole(
     }
     // check if the operation completed successfully and in case update the configuration
     if (isOperationSuccessful(resp.error_tag)) {
-      if (auto d = core_->get_cubes_dump()) {
-        d->UpdateCubesConfig(request.resource(), jbody, keys, op,
+      if (configuration::config.getCubesDumpEnabled()) {
```
frisso: I would integrate this line with the previous (if openation is succesfully and configdumpenabled then...)

src/polycubed/src/server/Resources/Endpoint/ParentResource.cpp
```diff
@@ -131,12 +131,12 @@ void ParentResource::CreateReplaceUpdate(
     }
     // check if the operation completed successfully and in case update the configuration
     if (isOperationSuccessful(resp.error_tag)) {
-      if (auto d = core_->get_cubes_dump()) {
+      if (configuration::config.getCubesDumpEnabled()) {
```
frisso: I would integrate the two 'if' in one line ("if A and B then")

src/polycubed/src/server/Resources/Endpoint/ParentResource.cpp
```diff
@@ -225,8 +225,8 @@ void ParentResource::del(const Request &request, ResponseWriter response) {
         std::vector<Response>{{ErrorTag::kNoContent, nullptr}},
         std::move(response));
     errors.push_back({ErrorTag::kCreated, nullptr});
-    if (auto d = core_->get_cubes_dump()) {
-      d->UpdateCubesConfig(request.resource(), nullptr, keys,
+    if (configuration::config.getCubesDumpEnabled()) {
+      core_->get_cubes_dump()->UpdateCubesConfig(request.resource(), nullptr, keys,
```
frisso: Just a question: there was a loop before, with possible multiple cubes returned.
Why this loop has been removed here? Do we have always one cube?

PR #186, opened 2019-07-25
libpolycube: rename functions in the utils library

francescomessina:
-ip_string_to_be_uint => ip_string_to_nbo_uint
-be_uint_to_ip_string => nbo_uint_to_ip_string

-mac_string_to_be_uint => mac_string_to_nbo_uint
-be_uint_to_mac_string => nbo_uint_to_mac_string

Where "nbo" means network byte order

### DIFF
--- a/src/libs/polycube/include/polycube/services/utils.h
+++ b/src/libs/polycube/include/polycube/services/utils.h
@@ -28,13 +28,15 @@ namespace polycube {
 namespace service {
 namespace utils {
 
-/* ip (a.b.c.d) to string and viceversa in big endian */
-uint32_t ip_string_to_be_uint(const std::string &ip);
-std::string be_uint_to_ip_string(uint32_t ip);
+/* ip (a.b.c.d) to string and viceversa
+ * Number is in network byte order (nbo), i.e., big endian */
+uint32_t ip_string_to_nbo_uint(const std::string &ip);
+std::string nbo_uint_to_ip_string(uint32_t ip);
 
-/* mac (aa:bb:cc:dd:ee:ff) to string and vicersa in big endian */
-uint64_t mac_string_to_be_uint(const std::string &mac);
-std::string be_uint_to_mac_string(uint64_t mac);
+/* mac (aa:bb:cc:dd:ee:ff) to string and vicersa
+ * Number is in network byte order (nbo), i.e., big endian */
+uint64_t mac_string_to_nbo_uint(const std::string &mac);
+std::string nbo_uint_to_mac_string(uint64_t mac);
 
 /* transforms an ipv4 dotted representation into a hexadecimal big endian */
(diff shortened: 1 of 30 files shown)


src/libs/polycube/include/polycube/services/utils.h
```diff
@@ -28,13 +28,15 @@ namespace polycube {
...
-/* ip (a.b.c.d) to string and viceversa in big endian */
-uint32_t ip_string_to_be_uint(const std::string &ip);
-std::string be_uint_to_ip_string(uint32_t ip);
-
-/* mac (aa:bb:cc:dd:ee:ff) to string and vicersa in big endian */
-uint64_t mac_string_to_be_uint(const std::string &mac);
-std::string be_uint_to_mac_string(uint64_t mac);
+/* ip (a.b.c.d) to string and viceversa in big endian
+ * nbo = network byte order */
+uint32_t ip_string_to_nbo_uint(const std::string &ip);
```
goldenrye: why we still keep 'big endian' keyword here? does endianness matter here?

src/libs/polycube/include/polycube/services/utils.h
```diff
@@ -28,13 +28,15 @@ namespace polycube {
...
+std::string nbo_uint_to_ip_string(uint32_t ip);
+
+/* mac (aa:bb:cc:dd:ee:ff) to string and vicersa in big endian
+ * nbo = network byte order */
```
frisso: ==>
/* mac (aa:bb:cc:dd:ee:ff) to string and vice versa
* Number is in network byte order (nbo), i.e., big endian */

mauriciovasquezbernal:
Just a question, what is the difference between "big endian" and "network byte order", can't both be used without changing the meaning?

goldenrye:
My understanding is that big-endian is a implementation how a number stored in computer while network order is used to transfer the number in the wire. But after doing some research, those two terms seem interchangeable. BTW not all the protocols use network order to transfer the data over the wire.

frisso:
Although big endian and network byte order are actually the same thing, conceptually are different. So, network byte order is the standard way to send data on wire. Big endian is one of the options we have to store data in CPUs.
I prefer "nbo" as we're getting a number from the wire, and we have to transform into a string or vice versa, *by keeping the same byte ordering we have on the wire*. So, "nbo" looks more appropriate to me, as we're keeping the same byte ordering we had on the wire.

mauriciovasquezbernal:
Thanks for the explanation to both of you.

frisso:
@francescomessina Can you please rebase? So that I can merge your PR. Thanks!

francescomessina:
> @francescomessina Can you please rebase? So that I can merge your PR. Thanks!

Done @frisso

PR #188, opened 2019-07-25
Print cleanup

frisso:
Added a void line at the end when printing the help on screen.
Updated the description of iptables services (it was void before) and adding standard metadata.

### DIFF
--- a/src/services/pcn-bridge/src/Fdb.cpp
+++ b/src/services/pcn-bridge/src/Fdb.cpp
@@ -18,7 +18,7 @@
 #include "Bridge.h"
 
 Fdb::Fdb(Bridge &parent, const FdbJsonObject &conf) : FdbBase(parent) {
-  logger()->info("Creating Fdb instance");
+  logger()->debug("[Fdb] Creating instance");
 
   agingTime_ = conf.getAgingTime();
 
@@ -28,7 +28,7 @@ Fdb::Fdb(Bridge &parent, const FdbJsonObject &conf) : FdbBase(parent) {
 }
 
 Fdb::~Fdb() {
-  logger()->info("Destroying Fdb instance");
+  logger()->debug("[Fdb] Destroying instance");
 }
 
 uint32_t Fdb::getAgingTime() {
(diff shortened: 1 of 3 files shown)


src/services/pcn-bridge/src/Fdb.cpp
```diff
@@ -18,7 +18,7 @@
 #include "Bridge.h"
 
 Fdb::Fdb(Bridge &parent, const FdbJsonObject &conf) : FdbBase(parent) {
-  logger()->info("Creating Fdb instance");
+  logger()->debug("[Fdb] Creating instance");
```
goldenrye: BTW do we support logging based on sub-category like enable only "Fdb"?

src/services/pcn-iptables/datamodel/iptables.yang
```diff
@@ -11,6 +11,12 @@ module iptables {
 
   uses "polycube-standard-base:standard-base-yang-module";
 
+  polycube-base:service-description "Iptables-compatible Service";
+  polycube-base:service-version "2.0";
+  polycube-base:service-name "iptables";
+  polycube-base:service-min-kernel-version "4.14.0";
```
acloudiator: Isn't the min-kernel-version be "4.15.0"?

https://github.com/polycube-network/polycube/blob/master/Documentation/installation.rst#updating-linux-kernel-to-v415

src/services/pcn-bridge/src/Fdb.cpp
frisso: It would be nice, but the current logging infrastructure doesn't support this for now.

src/services/pcn-iptables/datamodel/iptables.yang
frisso: The min kernel version in the YANG refers to the one that is actually required to run this service. The link you point out is the minimum version required to run the entire framework, which is actually max(version_required_by_each_individual_service).
So, this should stay 4.14 if I'm not mistaken.
@mbertrone : do you confirm?

PR #189, opened 2019-07-25
Updating name and documentation for Transparent HelloWorld Service

frisso:
Added documentation, useful for new developers.
It would be nice to shorten the name, so that we can avoid typing "transparenthelloworld" such as in;
   polycubectl transparenthelloworld add trhw0
but I'm unable to do so right now.


frisso:
Let's try; I remember something like the "-" character creates troubles with the code generation :-(

goldenrye:
> Let's try; I remember something like the "-" character creates troubles with the code generation :-(

Yes looks swagger auto-gem compiler has problem to generate correct code if there is "-" character, the character following the "-" will be used to generate variable but some place capital some place small-case and code won't be compiled.

src/services/pcn-transparent-helloworld/datamodel/transparenthelloworld.yang
```diff
@@ -23,7 +23,7 @@ module transparenthelloworld {
       enum SLOWPATH;
     }
     default PASS;
-    description "Action performed on ingress packets";
+    description "Action performed on ingress packets (DROP/PASS/SLOWPATH; default: PASS)";
```
mauriciovasquezbernal: I think information about possible values and default should not be part of the description.
Maybe the help generation could be improved to generate this.

mauriciovasquezbernal:
Few comments:
- `-` causes problems in the code generation. Infact the duplicated file you removed because of this.
- I don't know if this is a good idea to change the name on the datamodel without regenerating the code.
- Long names should not be a problem with bash autocompletion, you don't have to type all of them.

src/services/pcn-transparent-helloworld/datamodel/transparenthelloworld.yang
frisso: Right point, Mauricio. Moved now in the "example" section of the YANG. However, the example is not printed on screen by the help, so we should probably modify the command line to include that as well.

frisso:
> * Long names should not be a problem with bash autocompletion, you don't have to type all of them.

Unfortunately this is not true. Long names are consuming a lot of space on the screen.
Type for instance `polycubectl -h` and see how columns are large, due to the bit name we have.
So, I feel we should use shorter names for better visibility.

mauriciovasquezbernal:
> > * Long names should not be a problem with bash autocompletion, you don't have to type all of them.
> 
> Unfortunately this is not true. Long names are consuming a lot of space on the screen.
> Type for instance `polycubectl -h` and see how columns are large, due to the big name we have.
> So, I feel we should use shorter names for better visibility.

I soft agree with that, just find a name that is shorter but still meaningful, I think `tr-helloworld` is a quite good one as suggested above.

PR #190, opened 2019-07-26
Add commands to update git source code in install script

frisso:


mauriciovasquezbernal:
I am not sure about this change, why should this script update the source code?, what about if you want to install an old version?, this would be not possible with this as this pulls from git.

Same applies for docker and local compilation.

PR #196, opened 2019-07-28
Added option to update source code in install script

frisso:
Given the commits on my previous PR on this matter, I removed that PR and opened a new one.
In this case I kept everything as is, except that I added a new flag in install script so that you can (optionally) also update the source code in the repository. This enables to use this install script against with the most updated source code, while keeping the current behaviour of the script (i.e., install only) unchanged.
This should solve the concerts I've got on the previous PR.

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -33,9 +33,11 @@ Polycube installation script
 
 mode:
   pcn-iptables: install only pcn-iptables service and related components
-  pcn-k8s: install only pcn-k8s (Only to be used with docker build)
+  pcn-k8s: install only pcn-k8s (only to be used with docker build)
+  update: updates the source code of polycube before *default* install
   default: install all available polycube services"
 echo "$usage"
+echo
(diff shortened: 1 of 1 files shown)


mauriciovasquezbernal:
I am not completely sure `update` is a `mode`, for instance what about if you want to `update` and just install `pcn-iptables`?

Anyway this looks like a good solution for the time being for me.

frisso:
> I am not completely sure `update` is a `mode`, for instance what about if you want to `update` and just install `pcn-iptables`?

Mauricio, I thought about this, and I decided to take the easy way. Supporting two options, i.e., something like "update pcn-iptables" would complicate too much the script and currently nobody uses that feature.
So, I added some comments and printf to make clear what "update" does. If we need something more sophisticated, we'll go deeper and change again this script. Thanks for your comments (also in the previous PR), @mauriciovasquezbernal and @mbertrone !

Issue #198, opened 2019-07-29
polycubectl not installed in the docker image


goldenrye:
## Describe the bug
The pcn_k8s docker image generated doesn't contain the polycubectl binary, and user is not able to access cube status/statistics etc. information. Additionally after K8S security feature is introduced user has no way to check the firewall rule lists.

## To Reproduce
1. Go to polycube top directory
2. Excute the following commands to generate the pcn_k8s docker image
    # export DOCKER_BUILDKIT=1
    # docker build --build-arg MODE=pcn-k8s -t name:tag
3. no polycubectl binary generated

## Expected behavior
polycubectl binary should be in the /usr/local/bin in pcn_k8s docker image

## Please tell us about your environment: 
1. OS details: Ubuntu 18.04.1 LTS
2. Kernel details: Linux polycube-dev 4.19.0-041900-generic
3. Polycube Version: version: v0.1.0-beta+ [git: (branch/commit): master/d3034620-dirty]
4. Polycube Log: Attach the log from `/var/log/polycube/polycubed.log`

## Additional context
N/A

PR #199, opened 2019-07-29
generate the polycubectl binary for pcn_k8s docker image

goldenrye:
By default there is no polycubectl binary included in our official published pcn-k8s docker image: https://hub.docker.com/r/polycubenetwork/k8s-pod-network. We need this binary to help user to get more information of cubes running in the docker.


acloudiator:
@goldenrye Could you please add some description to this PR?! Thanks!

goldenrye:
> @goldenrye Could you please add some description to this PR?! Thanks!

Didn't realize there is no description, just add it and thanks for the reminding.

Dockerfile
```diff
@@ -3,7 +3,7 @@ FROM ubuntu:18.04
 ARG MODE=default
 RUN --mount=target=/polycube cp -r /polycube /tmp/polycube && \
 cd /tmp/polycube && \
-SUDO="" WORKDIR="/tmp/dev" ./scripts/install.sh $MODE && \
+SUDO="" USER="root" WORKDIR="/tmp/dev" ./scripts/install.sh $MODE && \
```
acloudiator: Do we need to specify the user as "root"? If I'm not mistaken as long as the user have sudo privilege, it should be fine. 
@mauriciovasquezbernal any thoughts?

acloudiator:
> By default there is no polycubectl binary included in our official published pcn-k8s docker image: https://hub.docker.com/r/polycubenetwork/k8s-pod-network. We need this binary to help user to get more information of cubes running in the docker.

@goldenrye Why do we want this inside the k8s-pod-network, isn't the `polycubectl` under "polycubenetwork/polycube" image.

Try this and let us know if this what you want.
```
sudo docker run -it polycubenetwork/polycube /usr/local/bin/polycubectl
```

goldenrye:
> > By default there is no polycubectl binary included in our official published pcn-k8s docker image: https://hub.docker.com/r/polycubenetwork/k8s-pod-network. We need this binary to help user to get more information of cubes running in the docker.
> 
> @goldenrye Why do we want this inside the k8s-pod-network, isn't the `polycubectl` under "polycubenetwork/polycube" image.
> 
> Try this and let us know if this what you want.
> 
> ```
> sudo docker run -it polycubenetwork/polycube /usr/local/bin/polycubectl
> ```

This command works but need add option "--net=host", but you have to run it in each node and won't be able to do that via kubectl like,
# kubectl exec polycube-sddh5 -n kube-system polycubectl show cubes 

polycube-sddh5 is a pod running in a slave node

Dockerfile
goldenrye: the problem here is that without setting the USER environment variable, docker building will fail in this line silently, 
install(CODE
"execute_process( COMMAND /bin/sh -c \"mkdir -p /var/log/polycube && chown $ENV{USER} /var/log/polycube \")")

frisso:
This relates to bug report #198.

SunSince90:
Hi, in order to use `polycubectl` and check firewall lists I have always followed the quickstart part of the documentation on a worker node (which basically is @acloudiator's command):

```
docker pull polycubenetwork/polycube:latest
docker run  -it --rm --privileged --network host \
-v /lib/modules:/lib/modules:ro -v /usr/src:/usr/src:ro -v /etc/localtime:/etc/localtime:ro \
polycubenetwork/polycube:latest /bin/bash -c 'polycubed -d && /bin/bash'
```
It always worked fine. 
Anyway, I see that entering inside a polycube pod like
`kubectl exec -it polycube-pfzjn -n kube-system -c polycubed -- /bin/bash`
and interacting with the polycubed of that node without leaving the k8s master could be a more convenient way than ssh-ing into the node and running a docker inside it.

frisso:
To me this PR makes sense.
Just a question: is this causing the size of our dockers to grow? To me, this size is already a little bit big, so I would avoid further increases if not strictly needed.

mauriciovasquezbernal:
I didn't enable this at that time because we're worried about the size of the image. I don't think this is an issue anymore, also the `polycubectl` size is about 10MB, what I think is negligible at this point.

frisso:
@goldenrye It looks there's a consensus to add this PR. I would suggest to add a brief text in the "troubleshooting" section here, so that people are aware of this possibility:
https://github.com/polycube-network/polycube/blob/master/Documentation/components/k8s/pcn-kubernetes.rst

PR #203, opened 2019-08-02
pcn-transparenthelloworld: removed the override of attach method.

francescomessina:
It is no longer necessary to run the 'get_parent_parameter("peer")' in the transparenthelloworld service, this generated a warning on the master branch


src/services/pcn-transparent-helloworld/src/Transparenthelloworld.cpp
```diff
@@ -70,16 +70,6 @@ void Transparenthelloworld::packet_in(polycube::service::Sense sense,
   }
 }
 
-void Transparenthelloworld::attach() {
```
frisso: Francesco, not sure we have to *delete* the attach method. To me, attach() makes sense.
What's wrong here is that we ask for the "peer" parameter, which clearly doesn't exist here.
Instead, we should ask for the interface we're attached to, and maybe log it, just to confirm that everything has gone well.

goldenrye: Agree with Fulvio, we should have a different implementation for such transparent cube (it is a service cube rather than transparent cube to be more accurate like pcn-synflood), just log and don't have to get the peer parameter.

PR #204, opened 2019-08-05
libpolycube: add a new parameter to the get_parameter in the extiface

francescomessina:
When a transparent service is connected to a port it asks for the name of the cube to which it has been connected. 
In the case in which a transparent service is attached to a netdev the name of the netdev is returned.
This PR adds a new parameter to the extiface class, which implements this behavior.

### DIFF
--- a/src/polycubed/src/extiface.cpp
+++ b/src/polycubed/src/extiface.cpp
@@ -30,6 +30,7 @@ namespace polycubed {
 
 const std::string PARAMETER_MAC = "MAC";
 const std::string PARAMETER_IP = "IP";
+const std::string PARAMETER_PEER = "PEER";
 
 std::set<std::string> ExtIface::used_ifaces;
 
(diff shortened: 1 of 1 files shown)


src/polycubed/src/extiface.cpp
```diff
@@ -372,6 +373,8 @@ std::string ExtIface::get_parameter(const std::string &param_name) {
       // netdev does not have an Ip
       return "";
     }
+  } else if (param_upper == PARAMETER_NAME) {
```
goldenrye: I think it is better than removing the API from the transparent_cube, but please remember to change the caller to get_parent_parameter(PARAMETER_NAME).

src/polycubed/src/extiface.cpp
```diff
@@ -30,6 +30,7 @@ namespace polycubed {
 
 const std::string PARAMETER_MAC = "MAC";
 const std::string PARAMETER_IP = "IP";
+const std::string PARAMETER_NAME = "PEER";
```
mauriciovasquezbernal: Shouldn't it be `PARAMETER_PEER`?

frisso:
@goldenrye Are you ok with this PR?

Issue #210, opened 2019-09-12
pcn_pkt_controller_with_metadata_stack() drops the packet


goldenrye:
## Describe the bug
pcn_pkt_controller_with_metada_stack() API is supposed to send the packet to both the controller and stack, but in existing implementation the packet was only sent to the controller and packet was dropped in stack.

## To Reproduce
1. Go to src/services/pcn-transparent-helloworld/src
2. Modify file TransparentHelloworld_dp_ingress.c replacing pcn_pkt_controller with pcn_pkt_controller_with_metadata_stack(), you need add one more parameter: uint32_t metadata[3] to make it compilable
3. compile the code and execute the polycubed, create a transparent hello world cube, set the cube to slow path action and attach the cube to a interface
4. before attach ping the interface is working, after attaching ping doesn't work

## Expected behavior
after attach the cube the interface, ping should still works

## Please tell us about your environment: 
1. OS details: Ubuntu 18.04.1 LTS 
2. Kernel details: 5.2.0-050200-generic
3. Polycube Version: b84e3b2f

## Additional context
the problem is that to_stack flag carried by skb->cb[0] was overwritten when handle_rx returns RX_CONTROLLER,
    case RX_CONTROLLER:
      skb->cb[0] = CUBE_ID | TYPE << 16;
      return to_controller(skb, md.reason);

PR #211, opened 2019-09-12
keep the pass_to_stack flag in skb 

goldenrye:
some caller like pcn_pkt_controller_with_metadata_stack may set this flag. pass_to_stack flag was carried by cb[0] in the first MSB in the last 16 bits (0x8000) and this flag will be used to decide if the packet should be sent to stack for further processing or dropped. 

this fix issue #210

### DIFF
--- a/src/polycubed/src/transparent_cube_tc.cpp
+++ b/src/polycubed/src/transparent_cube_tc.cpp
@@ -99,7 +99,7 @@ int handle_rx_wrapper(struct CTXTYPE *skb) {
     case RX_DROP:
       return TC_ACT_SHOT;
     case RX_CONTROLLER:
-      skb->cb[0] = CUBE_ID | TYPE << 16;
+      skb->cb[0] = (skb->cb[0] & 0x8000) | CUBE_ID | TYPE << 16;
       return to_controller(skb, md.reason);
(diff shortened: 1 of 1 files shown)


frisso:
To me looks perfect. @mauriciovasquezbernal , do you seen any problem? Thanks!

src/polycubed/src/transparent_cube_tc.cpp
```diff
@@ -99,7 +99,7 @@ int handle_rx_wrapper(struct CTXTYPE *skb) {
     case RX_DROP:
       return TC_ACT_SHOT;
     case RX_CONTROLLER:
-      skb->cb[0] = CUBE_ID | TYPE << 16;
+      skb->cb[0] = skb->cb[0] & 0x8000 ? (CUBE_ID | TYPE << 16 | 0x8000) : (CUBE_ID | TYPE << 16);
```
mauriciovasquezbernal: Could `(skb->cb[0] & 0x8000) | CUBE_ID | TYPE << 16` work as well?

goldenrye: 👍 yes, this expression is more concise than the original one. I have revised it.

PR #214, opened 2019-09-20
pcn-k8s: Fix pcn-k8s compatibility issue with 1.16

SunSince90:
The newest version of Kubernetes, 1.16, has dropped support for resources that use `apiVersion: extensions/v1beta1`: this made `pcn-k8s` incompatible with Kubernetes 1.16, as some of its resources rely on that version.

This PR fixes this compatibility issue and updates documentation to specify the minimum Kubernetes version supported by pcn-k8s.

### DIFF
--- a/Documentation/components/k8s/pcn-kubernetes.rst
+++ b/Documentation/components/k8s/pcn-kubernetes.rst
@@ -318,3 +318,8 @@ Developing
 ----------
 
 Refer to :doc:`Developers <developers>`
+
+Compatibility
+----------
+
+Pcn-k8s is compatible with all versions equal or greater than 1.9, although we recommend the latest version of Kubernetes.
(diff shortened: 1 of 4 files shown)


Documentation/components/k8s/pcn-kubernetes.rst
```diff
@@ -318,3 +318,8 @@ Developing
 ----------
 
 Refer to :doc:`Developers <developers>`
+
+Compatibility
+----------
+
+Although being on the latest version of Kubernetes is reccomended, pcn-k8s is compatible with all versions equal or greater than 1.9.
```
frisso: ==>
Pcn-k8s is compatible with all versions equal or greater than 1.9, although we recommend the latest version of Kubernetes.

SunSince90: Just updated, thank you.

Issue #217, opened 2019-09-26
packet was dropped when transparent cube sends packet to slow path


goldenrye:
## Describe the bug
Attaching a transparent cube to a interface and if the cube's egress handler has to send packet to the slow path and then back to fast path, existing framework doesn't work. There are a few issues in exiting implementation: polycubed cores dump, packet dropped and dead loop happens .

## To Reproduce
1. Modify the transparent hello world cube packet_in() to call send_packet_out(), this expects the packet sent back to fast path without doing anything
2. Create a cube and configure the cube ingress action to "pass", egress action to "slowpath", and attach the cube to a interface
3. Send ICMP echo request from remote host to the interface
4. ICMP echo reply won't be able to send back to remote host

## Expected behavior
Attaching interface to helloworld cube shouldn't drop the packet

## Please tell us about your environment: 
1. OS details: Ubuntu 18.04.1 LTS
2. Kernel details: Linux 4.19.56
3. Polycube Version: master/f4dd91d7

## Additional context
[Jianwen Pi] 
@mauriciovasquezbernal In existing design, the control plane builds the topology (NH chain) for cubes when transparent cubes were attached to a interface. when a cube decides to forward a packet to slow path, the packet will be forwarded to controller and come back to the cube itself or next cube depending on the parameter of API send_packet_out(). No matter if a packet is sent to the slow path in ingress or egress TC hook, the packet will be injected into the ingress context via pcn_tc_cp interface, the 2 diagram I present in the doc is trying to illustrate the packet path in different scenarioes.

[mauriciovasquezbernal]
So basically the egress is broken
the packet out in egress is broken, because the pcn pass is not able to send it in egress actually.
I'll go a step back first, do you have a use case in mind for this?

[Jianwen Pi]
yes, the egress is broken. Even for ingress if the packet is coming back to itself there is still issues. I don’t have solid use case, but as framework we have to support this or being able to detect the loop and drop the packet. BTW in controller fast path code, there is no code to handle the stack case.

[mauriciovasquezbernal]
ok, got it.
Unfortunately I don't have answer to all this points, the slow path support in transparent cubes is quite primitive as you already found. Honestly I don't know if this is a good idea to invest time on those cases or just document them as issues.

PR #218, opened 2019-09-26
Fix some issues for transparent cube when packet forwarded to slow path 

goldenrye:
The modification fixed the following issues:
  - core dump when transparent cube sends packet to the slow path
  - stack topology (0xffff) not recognized by controller bpf program
  - calling send_packet_out() in transprent cube slow path may cause dead loop

Issue #217


src/polycubed/src/controller.cpp
```diff
@@ -136,8 +141,19 @@ int controller_module_rx(struct __sk_buff *ctx) {
...
+    u32 flags = md->flags;
 
 	ctx->cb[0] = in_port << 16 | module_index;
+    ctx->cb[2] = flags;
+    if (module_index == 0xffff) {
+        pcn_log(ctx, LOG_INFO, "[tc-decapsulator]: NH is stack, flags: 0x%x", flags);
+        if (flags & MD_EGRESS_CONTEXT) {
+            return bpf_redirect(in_port, 0);
+        } else {
+            return 0;
```
mauriciovasquezbernal: What is `return 0` for in this case?

src/polycubed/src/transparent_cube.cpp
```diff
@@ -99,21 +99,33 @@ void TransparentCube::send_packet_out(const std::vector<uint8_t> &packet,
 
   uint16_t port = 0;
```
mauriciovasquezbernal: Not a big deal but recirculate support is still missing.

src/polycubed/src/controller.cpp
```diff
@@ -445,7 +447,18 @@ void Controller::send_packet_to_cube(uint16_t module_index, uint16_t port_id,
   }
   metadata_table_->update_value(ctrl_rx_md_index_, md_temp);
 
-  iface_->send(const_cast<std::vector<uint8_t> &>(packet));
+  if (sense == service::Sense::EGRESS) {
+      iface_->send(const_cast<std::vector<uint8_t> &>(packet));
+  } else {
+      /* ingress packet needs modifying the destination MAC address
```
mauriciovasquezbernal: I don't totally agree with this change as it will limit the packets that a transparent cube is able to inject in the fast path. I think the best approach is not to change it and let it under the responsibility of the user.

mauriciovasquezbernal:
According to my understanding this PR is mainly focused in the case a transparent cube is attached to a NIC, and the support when the transparent cube is attached to another cube's port is not changed with this PR, right?

Overall looks good to me, maybe clarifying the previous point will help to understand it better.

frisso:
Another comment, Michel. Can you please rebase against master? It makes the commit more clean.

src/polycubed/src/controller.cpp
goldenrye: return 0 (TC_ACK_OK) means the packet will pass to Linux stack. For packet forwarded to slow path in ingress context should continue the stack process (routing, filter, egress BPF etc.) before it was sent out.

src/polycubed/src/controller.cpp
goldenrye: A packet received from a physical NIC has destination MAC to this NIC, when it was forwarded to the slow path and injected back to pcn_tc_cp interface, the packet will be dropped by Linux kernel because the packet doesn't have a valid MAC address for IP processing. In my testing the packet will be classified as PACKET_OTHERS and dropped before calling the ip_rcv().

goldenrye:
> According to my understanding this PR is mainly focused in the case a transparent cube is attached to a NIC, and the support when the transparent cube is attached to another cube's port is not changed with this PR, right?
> 
> Overall looks good to me, maybe clarifying the previous point will help to understand it better.

@mauriciovasquezbernal  Yes this PR is only focusing on the transparent cube, it shouldn't change any behavior for regular cube.

src/polycubed/src/controller.cpp
mauriciovasquezbernal: It makes sense. There only one point I am not totally sure. This eBPF program is executed in the context of the `pcn_tc_cp` ingress hook, then the OS will think the packet is coming from that interface. Do you think this is a problem?

src/polycubed/src/controller.cpp
mauriciovasquezbernal: I think this is related to my previous comment, the OS discards the packet because it appears to be coming from the `pcn_tc_cp` interface.  If there is not another solution for this problem I'll ask to scope this modification only to the case a transparent service is attached to a NIC.

mauriciovasquezbernal:
It could be a good idea to extend the description of the PR making it clear that it adds support for the case a transparent cube is attached to a NIC.

For the rest, it LGMT.

src/polycubed/src/controller.cpp
goldenrye: The ingress interface is not determined by the interface hook where BPF program is executed (bpf_call can change the context at any time), the interface is determined by the interface driver that picking up the packet, which in our case pcn_tc_cp when packet re-injected to this interface by calling send_packet_to_cube(). But recall that the packet is originally arriving in another interface (with dest MAC to that interface) before it was sent to slow path, without modifying the dest MAC kernel wouldn't happy with that.

src/polycubed/src/controller.cpp
goldenrye: Sure this is a good point I think we should make this change to transparent scope. Thanks for pointing out this.

src/polycubed/src/controller.cpp
mauriciovasquezbernal: I'm not totally sure I understood your comment.
Let's think about this case, a transparent cube is attached to eth0.
A packet arrives to `eth0`, it is sent to the slowpath and then the slow paths performs a `packet_out` to send it to the linux stack. With this architecture the packet will be passed to the linux stack from the `pcn_tc_cp` interface and not from `eth0`.  Is this is a problem besides the MAC address?

src/polycubed/src/controller.cpp
```diff
@@ -418,14 +436,30 @@ void Controller::unregister_cb(int id) {
...
+  if (sense == service::Sense::EGRESS) {
+      md_temp.flags |= MD_EGRESS_CONTEXT;
+  }
   metadata_table_->update_value(ctrl_rx_md_index_, md_temp);
 
-  iface_->send(const_cast<std::vector<uint8_t> &>(packet));
+  if (!is_tr_cube || sense == service::Sense::EGRESS) {
+      iface_->send(const_cast<std::vector<uint8_t> &>(packet));
+  } else {
+      /* for transparent cube ingress packet needs modifying the
```
mauriciovasquezbernal: Should the MAC also be changed when the transparent cube is attached to another cube's port?

mauriciovasquezbernal:
Is XDP support missing from this PR?, will be done in the future or it's not needed?

src/polycubed/src/controller.cpp
goldenrye: I didn't test the regular cube but I think the MAC shouldn't impact the regular cube processing because there is no chance the packet back to kernel stack in regular cube case.

goldenrye:
> Is XDP support missing from this PR?, will be done in the future or it's not needed?

You are right, this PR doesn't cover XDP case. Will evaluate whether XDP is impacted or not in the future.

src/polycubed/src/controller.cpp
mauriciovasquezbernal: Actually I am not worried about the stack case but more about breaking standard cubes. What about if the standard cube uses the MAC of the packet to do any calculation?.

goldenrye: Oh, I understood your point and modified the code limiting the scope to transparent && ingress context && attached to a interface.

PR #219, opened 2019-10-03
Updating libyang-dev manual installation commands

michelsciortino:
This PR updates the list of commands needed to install libyang-dev present in the Polycube manual installation guide to align them with those present in the automatic installation script.

### DIFF
--- a/Documentation/installation.rst
+++ b/Documentation/installation.rst
@@ -102,11 +102,12 @@ Install libyang-dev
 
 ::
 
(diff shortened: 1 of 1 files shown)


frisso:
Michel, do you have time to try if the new script still works on 16.04?
Thanks!

michelsciortino:
> Does this mean we have to use Ubuntu18.04 for development environment, or it still support 16.04?

As the Documentation says "It is not guaranteed that the script works on versions or distributions other than Ubuntu 18.04".
In fact, the script does not work on `Ubuntu 16.04` as it comes with a version of cmake and g++ that are not compatible with C++ 17;
Upgrading cmake to version 3.10.2 and g++ to version 7 it is possible to use the script to build and install Polycube on `Ubuntu 16.04`.

PR #220, opened 2019-10-03
Removing references to polycube-tools

michelsciortino:
This PR removes all the references to polycube-tools from source code and documentation.


frisso:
LGTM.
Just add some more context in the commit description: "polycube tools were used to capture and dump packets in Polycube, e.g., for debugging purposes. Now this feature is already part of the framework, hence we can remove this dependency."

PR #222, opened 2019-10-15
Packetcapture service

DavideAG:
Thanks to this pull request it will be possible to use the transparent "pcn-packetcapture" service in polycube.
This service allows you to capture packets passing through an interface.
A demonstration client using the REST API has been added to the service directory. For more information, consult the documentation (https://github.com/polycube-network/polycube/commit/d4be2d987e1097694f2ee3b93079b9673d0b9604)


src/services/pcn-packetcapture/src/Packetcapture_dp_egress.c
```diff
@@ -0,0 +1,198 @@
...
+      }else if((filters_tab->l4proto_filter == 2) && (ip->protocol != IPPROTO_UDP)){   /* udp only */
+        return RX_OK;
+      }
+    }
+
+    pkt->ip = 1;
+    pkt->srcIp = ip->saddr;
+    pkt->dstIp = ip->daddr;
+    if (ip->protocol == IPPROTO_TCP) {
+      tcp = data + sizeof(*ethernet) + sizeof(*ip);
```
goldenrye: Need to consider the case here when a packet contains the IP option

src/services/pcn-packetcapture/src/Packetcapture_dp_egress.c
```diff
@@ -0,0 +1,198 @@
...
+
+      /* dst port filter */
+      if((filters_tab->dst_port_flag == true) && (filters_tab->dst_port_filter != ntohs(tcp->dest))){
+        return RX_OK;
+      }
+
+      pkt->srcPort = tcp->source;
+      pkt->dstPort = tcp->dest;
+    } else if (ip->protocol == IPPROTO_UDP) {
+      udp = data + sizeof(*ethernet) + sizeof(*ip);
```
goldenrye: Same as above

src/services/pcn-packetcapture/src/Packetcapture_dp_egress.c
```diff
@@ -0,0 +1,198 @@
...
+      if (data + sizeof(*ethernet) + sizeof(*ip) + sizeof(*udp) > data_end)
+        return RX_DROP;
+      pkt->l4proto = IPPROTO_UDP;
+      pkt->srcPort = udp->source;
+      pkt->dstPort = udp->dest;
+    }
+
+  }
```
goldenrye: Looks you send all the traffic to the controller, shouldn't we let the non-IP pass through?

DavideAG: in this case we would lose the non-IP traffic in the capture, like ARP messages. 
I thought it useful not to discard that type of traffic

goldenrye: if we want to capture the non-ip packet, we should probably specify it in the filter by adding some protocol support like ARP(or at least we have some knob to collect all the non-IP protocol packet), otherwise we will get all the non-IP packet in controller, which I think bring a couples of issues: 1. performance; 2. break the protocol. Think about when you enable a interface packet capture feature if the interface has IPSEC traffic, OSPF, IGMP etc...

frisso:
Davide, you're missing to add your name in the AUTHORS file.

src/services/pcn-packetcapture/src/Filters.cpp
```diff
@@ -0,0 +1,158 @@
...
+void Filters::setSrc(const std::string &value) {
+  srcIp = value;
+
+  uint32_t ip_src_filter = 0;
+  netmaskSrc = (0xFFFFFFFF << (32 - std::stoi(value.substr(value.find("/")+1)))) & 0xFFFFFFFF;
+  std::string source_ip = value.substr(0, value.find("/"));
+  inet_pton(AF_INET, source_ip.data(), &ip_src_filter);
+  ip_src_filter = ntohl(ip_src_filter);
+  networkSrc = ip_src_filter & netmaskSrc;
+  //network_packet = pkt_values.srcIp & netmask_filter;     //TODO: do it in the fast path
```
frisso: Can this line be deleted?

src/services/pcn-packetcapture/src/Filters.cpp
```diff
@@ -0,0 +1,158 @@
...
+void Filters::setDst(const std::string &value) {
+  dstIp = value;
+
+  uint32_t ip_dst_filter = 0;
+  netmaskDst = (0xFFFFFFFF << (32 - std::stoi(value.substr(value.find("/")+1)))) & 0xFFFFFFFF;
+  std::string source_ip = value.substr(0, value.find("/"));
+  inet_pton(AF_INET, source_ip.data(), &ip_dst_filter);
+  ip_dst_filter = ntohl(ip_dst_filter);
+  networkDst = ip_dst_filter & netmaskDst;
+  //network_packet = pkt_values.dstIp & netmask_filter;     //TODO: do it in the fast path
```
frisso: Can this line be deleted?

src/services/pcn-packetcapture/src/Packetcapture.cpp
```diff
@@ -0,0 +1,358 @@
...
+    pcap_header->version_minor = global_header->getVersionMinor();
+    pcap_header->thiszone = global_header->getThiszone();   //timestamp are always in GMT
+    pcap_header->sigfigs = global_header->getSigfigs();
+    pcap_header->snaplen = filters->getSnaplen();
+    pcap_header->linktype = global_header->getLinktype();
+
+    myFile.write(reinterpret_cast<const char*>(pcap_header), sizeof(*pcap_header));
+    writeHeader = false;
+  }
+  gettimeofday(&tp, NULL);
```
frisso: If I understand correctly, we assign the timestamp to the packet in the control plane. This may not be correct, especially if we want to have a precise timestamping (e.g., in case of forensic investigations).
We should timestamp the packet while in dataplane, in order to be more precise, and send the packet in userspace including this number.

src/services/pcn-packetcapture/src/Packetcapture.cpp
```diff
@@ -0,0 +1,358 @@
...
+void Packetcapture::packet_in(polycube::service::Direction direction,
+    polycube::service::PacketInMetadata &md,
+    const std::vector<uint8_t> &packet) {
+  
+  Tins::EthernetII pkt(&packet[0], packet.size());
+      
+  switch (direction) {
+    case polycube::service::Direction::INGRESS:
+    if( getNetworkmode() == true ){
+      addPacket(packet);    /* store the packet */
```
frisso: "store the packet" means "Store packet in the FIFO buffer", correct?

src/services/pcn-packetcapture/src/Packetcapture_dp_egress.c
```diff
@@ -0,0 +1,163 @@
...
+
+    } else if (ip->protocol == IPPROTO_UDP) {
+      udp = data + sizeof(*ethernet) + iphdr_size;
+      if (data + sizeof(*ethernet) + iphdr_size + sizeof(*udp) > data_end)
+        return RX_DROP;
+    }
+
+  }
+  u16 reason = 1;
+  return pcn_pkt_controller(ctx, md, reason);
```
frisso: This introduces a big performance penalty, as all the traffic is sent to controller and then back here.
Please add a big "todo" here, suggesting an alternative implementation and telling people why you were forced to stay with this behavior.

src/services/pcn-packetcapture/src/Packetcapture_dp_egress.c
goldenrye: Ok, then we should add a feature improvement issue.

Documentation/services/pcn-packetcapture/Packetcapture.rst
```diff
@@ -0,0 +1,136 @@
...
+
+    #create the packetcapture service
+    polycubectl packetcapture add sniffer capture=bidirectional
+
+the packet capture service can capture packets in three differents ways:
+
+- ingress only: **capture=ingress**
+- egress only: **capture=egress**
+- bidirectional: **capture=bidirectional**
+- not working mode: **capture=off**
```
goldenrye: a little confusing: this is "networking mode" or "not working mode"? I see there is networking mode in later section.

src/services/pcn-packetcapture/src/Packetcapture.cpp
```diff
@@ -0,0 +1,363 @@
...
+    p->setTimestampSeconds((uint32_t) tp.tv_sec);
+    p->setTimestampMicroseconds((uint32_t) tp.tv_usec);
+    p->setPacketlen((uint32_t) packet.size());
+    if(p->getPacketlen() > filters->getSnaplen()){
+      p->setCapturelen(filters->getSnaplen());
+    }else{
+      p->setCapturelen(p->getPacketlen());
+    }
+    p->setRawPacketData(packet);
+    packets_captured.push_back(p);
```
goldenrye: just curious in networking mode if no client consumes the packet, how many packet will be captured in the list?

DavideAG:
> looks good fix to the commit [4b42cef](https://github.com/polycube-network/polycube/commit/4b42cefa9a04dde909f327eeb56f1b3d240ad175), one request is that can you squash this fix and commit [4b42cef](https://github.com/polycube-network/polycube/commit/4b42cefa9a04dde909f327eeb56f1b3d240ad175) to one commit so the original issue can be resolved?

@goldenrye Sure, I'm doing it!

src/services/pcn-packetcapture/src/Packetcapture.cpp
DavideAG: currently there is no maximum size that limits the amount of stored packets.
This can be a problem for the memory used. 

We can try solutions like:

- Temporarily dump packets in the file system
- use structures such as ring buffers, admitting to having to eventually have a maximum number of packages in memory

Documentation/services/pcn-packetcapture/Packetcapture.rst
DavideAG: Hi @goldenrye ,

this option indicates the direction of the packages that the service must capture.
The direction of the captured packets is independent of the operation in "nework mode" or non "network mode" of the service.


For example, we would like to be able to capture only incoming traffic to the service using the option:
_polycubectl sniffer set capture=ingress_

DavideAG:
> Very good work, @DavideAG
> Let's fix the minor issues pointed out in the review. Thanks!

@frisso,

Fine, I'm working to resolve this.

src/services/pcn-packetcapture/src/Packetcapture.cpp
DavideAG: That is correct, afther this instruction the packet will be stored in the FIFO queue

Documentation/services/pcn-packetcapture/Packetcapture.rst
goldenrye: so the line 38 should be "networking mode" instead of "not working mode", right?

frisso: Jianwen, actually not, line 38 is just for turning capture off. I modified the code now, please check if now it looks better.

DavideAG: @frisso Thank you for your commit

goldenrye:
Conditionally approved assuming the comments and change requested are fixed. Additionally there are some caveats and potential improvement should be planned for this feature.

src/services/pcn-packetcapture/src/Packetcapture.cpp
```diff
@@ -52,6 +56,14 @@ struct pcap_pkthdr {
 Packetcapture::Packetcapture(const std::string name, const PacketcaptureJsonObject &conf)
   : TransparentCube(conf.getBase(), { packetcapture_code_ingress }, { packetcapture_code_egress }),
     PacketcaptureBase(name) {
+  
+  gettimeofday(&ts, NULL);              //getting actual time
```
frisso: Davide, one more comment: can you please add a couple of sentences here to present how the timestamping algorithm works? It is not immediate while looking at the code.

src/services/pcn-packetcapture/src/Packetcapture.cpp
```diff
@@ -105,7 +116,17 @@ void Packetcapture::writeDump(const std::vector<uint8_t> &packet){
     myFile.write(reinterpret_cast<const char*>(pcap_header), sizeof(*pcap_header));
     writeHeader = false;
   }
-  gettimeofday(&tp, NULL);
+  struct timeval tp;
```
frisso: Add a comment such as "The algorithm for timestamping packets is described in file XX function YY"

src/services/pcn-packetcapture/src/Packetcapture_dp_egress.c
```diff
@@ -193,6 +198,13 @@ static __always_inline int handle_rx(struct CTXTYPE *ctx, struct pkt_metadata *m
 
   }
 
+  /* Getting packet timestamp */
```
frisso: Add a comment such as "The algorithm for timestamping packets is described in file XX function YY"

src/services/pcn-packetcapture/src/Packetcapture_dp_ingress.c
```diff
@@ -186,7 +191,14 @@ static __always_inline int handle_rx(struct CTXTYPE *ctx, struct pkt_metadata *m
     }
 
   }
-
+  
+  /* Getting packet timestamp */
+  uint64_t *pkt_timestamp = packet_timestamp.lookup(&key);
```
frisso: Add a comment such as "The algorithm for timestamping packets is described in file XX function YY"

src/services/pcn-packetcapture/src/Packetcapture.cpp
DavideAG: Sure!
I'm doing it

src/services/pcn-packetcapture/src/Packetcapture.cpp
```diff
@@ -0,0 +1,405 @@
...
+   * Now we are calculating the system start time in epoch.
+   * 'timeval struct ts' rapresents the system start time in epoch calculated as
+   * actual time - system uptime
+   * 
+   * See line 138 of this file for more details
+   */
+
+
+  gettimeofday(&ts, NULL);              //getting actual time
+  std::uint64_t uptime = std::chrono::duration_cast<std::chrono::nanoseconds>(std::chrono::high_resolution_clock::now().time_since_epoch()).count();
```
sebymiano: @DavideAG I might be wrong but, what is the different between `gettimeofday` and uptime?
It looks to me that they are just two different ways of getting the current time but the produced result should be the same.
Moreover, if I remember correctly, you want to subtract the uptime from the current time and then add the value that you get from the data path to retrieve the final time on which the packet was captured.
If I am not wrong, the way you are calculating the uptime is not correct (see [1]). Am I missing something?

[1] https://stackoverflow.com/questions/30095439/how-do-i-get-system-up-time-in-milliseconds-in-c

src/services/pcn-packetcapture/src/Packetcapture.cpp
```diff
@@ -0,0 +1,405 @@
...
+  struct timeval tp;
+  tp.tv_sec = ts.tv_sec;
+  tp.tv_usec = ts.tv_usec;
+  time_t sec_off = temp_offset/BILLION;     /* from nanoseconds to seconds */
+  temp_offset -= sec_off*BILLION;
+  tp.tv_usec += temp_offset/ONE_THOUSAND;   /* from nanoseconds to microseconds */
+  tp.tv_sec += sec_off;
+  sec_off = tp.tv_usec/MILLION;             /* from microseconds to seconds */
+  tp.tv_usec -= sec_off*MILLION;
+  tp.tv_sec += sec_off;
```
sebymiano: A general comment to this code here. Why don't you try to use more C++ features to make the code more readable?

For instance, instead of saving the `struct timeval ts` as a member of this class, you could save the C++ `timepoint` object; then after calculating the difference between the two objects you could extract the secs and microseconds that you need to insert into the pcap header.
This would make the code more readable and portable (even if it is not really needed in our case).

src/services/pcn-packetcapture/src/Packetcapture.cpp
DavideAG: Thank you for your comment @sebymiano !

I changed this part of the code and it will be fixed in the next commit.

src/services/pcn-packetcapture/src/Packetcapture.cpp
DavideAG: Well, I changed this part.
Thank you for your help @sebymiano

src/services/pcn-packetcapture/src/Packetcapture.cpp
```diff
@@ -0,0 +1,409 @@
...
+   */
+
+  struct timeval tp;
+  std::chrono::system_clock::time_point timeBase;
+  timeBase = timeP + temp_offset;
+  std::chrono::system_clock::duration tp_dur = timeBase.time_since_epoch();
+  
+  std::chrono::seconds s = std::chrono::duration_cast<std::chrono::seconds>(tp_dur);
+  tp_dur -= s;
+  std::chrono::microseconds us = std::chrono::duration_cast<std::chrono::microseconds>(tp_dur);
```
sebymiano: Definitely better! Thanks @DavideAG.
Just last comment (yes, I know, I am maybe too fussy).
Wouldn't it be better to create a function that does the conversion? For example
```
template<typename Duration>
void to_timeval(Duration&& d, struct timeval & tv) {
    std::chrono::seconds const sec = std::chrono::duration_cast<std::chrono::seconds>(d);

    tv.tv_sec  = sec.count();
    tv.tv_usec = std::chrono::duration_cast<std::chrono::microseconds>(d - sec).count();
}
```
In this way you will just call `to_timeval(timeBase, &tp)` and it's done ;)

DavideAG: Yes it's better using a function like this. Thank you for your help!

PR #223, opened 2019-10-16
Fixing jenkins pipeline & tests

michelsciortino:
Docker base image is now built only triggered by a `cron` at 2:00 am (everyday) to reduce the building time of the pipeline.
Mounting volumes of the polycubed image have been changed:
- `/proc` is now mounted under `/host/proc` to have a complete visibility of the namespaces created outside of the container.

Solving iptables tests:
Docker manipulates the iptables policy of the node during his installation and this makes the kernel drop the packets forwarded between different network namespaces; to solve this, `sudo iptables -P FORWARD ACCEPT` must be run on the host before running tests in order avoid the kernel drops packets.

### DIFF
--- a/CI/JenkinsfileTestCleanInstance
+++ b/CI/JenkinsfileTestCleanInstance
@@ -5,7 +5,7 @@ def slackNotification(buildStatus,String notification_url = "",project_name = ""
         title = "${BRANCH_NAME} - BUILD UNSTABLE on ${project_name}"
         text = "Build n°${BUILD_NUMBER} of branch ${BRANCH_NAME} unstable (${BUILD_URL})"
         colorCode = '#FFFF00'
-    } else if (buildStatus == 'SUCCESSFUL') {
+    } else if (buildStatus == 'SUCCESS') {
         text = "Build n°${BUILD_NUMBER} of branch ${BRANCH_NAME} successful (${BUILD_URL})"
         title = "${BRANCH_NAME} - BUILD SUCCESS on ${project_name}"
         colorCode = '#00FF00'
@@ -16,11 +16,11 @@ def slackNotification(buildStatus,String notification_url = "",project_name = ""
     } else {
         echo "The build status is: ${buildStatus}"
         title = "${BRANCH_NAME} - BUILD FAILED on ${project_name}"
-        text = "Build n°${BUILD_NUMBER} of branch ${BRANCH_NAME} failed (${BUILD_URL})"
+        text = "Build n°${BUILD_NUMBER} of branch ${BRANCH_NAME} failed with status ${buildStatus} (${BUILD_URL})"
         colorCode = '#FF0000'
     }
 
-def payload= JsonOutput.toJson(
+    def payload= JsonOutput.toJson(
         [
             attachments: [
                 [
@@ -34,7 +34,6 @@ def payload= JsonOutput.toJson(
 
     sh "curl -X POST --fail -H 'Content-type: application/json' --data '${payload}' ${notification_url}"
 }
-
 
 /**
 * Trigger Github PR Check ( can be : error, failure, pending, or success)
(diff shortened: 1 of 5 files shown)


CI/JenkinsfileTestCleanInstance
```diff
@@ -58,331 +57,348 @@ pipeline {
...
         }
-      }
-     post {
-        failure {
-          script {
-            setGithubStatus("${commit_id}", "pending", github_token, "polycube-network/polycube", "Docker Image Build", "Docker Image Build")
-          }
+        stage('Build Base Image') {
+            when {
+                triggeredBy 'TimerTrigger'
```
palexster: the branch should also be master otherwise it will built at 2am for every branch

CI/JenkinsfileTestCleanInstance
```diff
@@ -58,331 +57,348 @@ pipeline {
...
+                }
             }
         }
-      }
-     post {
-        failure {
-          script {
-            setGithubStatus("${commit_id}", "pending", github_token, "polycube-network/polycube", "Docker Image Build", "Docker Image Build")
-          }
+        stage('Build Base Image') {
```
palexster: This step was not skipped in the last pipeline

CI/JenkinsfileTestCleanInstance
```diff
@@ -58,331 +57,372 @@ pipeline {
...
+                label "docker"
+            }
+            steps {
+                script {
+                    docker.withRegistry("", 'polycube-repo') {
+                        sh """
+                            export DOCKER_BUILDKIT=1
+                            docker pull "polycubebot/${image_name}-default:${image_tag}"
+                            docker tag "polycubebot/${image_name}-default:${image_tag}" "polycube-network/polycube:latest"
+                            docker push "polycube-network/polycube:latest"
```
palexster: The dash between polycube and network should be removed

palexster: polycubenetwork is the right name

michelsciortino: Done.

Dockerfile
```diff
@@ -32,4 +32,4 @@ ADD src/components/k8s/cni/cni-uninstall.sh /cni-uninstall.sh
 ADD src/components/k8s/pcn_k8s/cleanup.sh /cleanup.sh
 ADD src/components/k8s/pcn_k8s/init.sh /init.sh
 
-CMD ["polycubed"]
+CMD ["nsenter","--mount=/host/proc/1/ns/mnt","polycubed"]
```
frisso: Can you please add a comment saying what is this about?

src/services/pcn-nat/test/test_tcp_snat.sh
```diff
@@ -16,9 +16,9 @@ source "${BASH_SOURCE%/*}/helpers.bash"
 
 function test_tcp {
     sudo ip netns exec ns2 netcat -l -w 5 $tcp_port&
-	sleep 2
+	sleep 5
```
frisso: Would it be possible to define a constant that is valid for all the tests, so that we don't have to change it manually throughout the entire file?

frisso:
@michelsciortino Thanks for this long awaited patch!
Two minor comments: can you add what you said in the commit description also in the proper location of the script files, so that we know immediately the reasons of the new commands?

     /proc is now mounted under /host/proc to have a complete visibility of the namespaces created outside of the container.

     sudo iptables -P FORWARD ACCEPT must be run before running test in order avoid the kernel drops

src/services/pcn-nat/test/test_tcp_snat.sh
michelsciortino: I think this should be done in another PR as this change should affect every service's tests.

frisso: Good point. Leave this as is.

Dockerfile
michelsciortino: Done.

michelsciortino:
>>  /proc is now mounted under /host/proc to have a complete visibility of the namespaces created outside of the container.
I'v added this as a comment in the jenkinksfile

>>  sudo iptables -P FORWARD ACCEPT must be run before running test in order avoid the kernel drops
This command is not present in the scripts as it must be run on the host;
I've added a better explanation on the PR description

PR #226, opened 2019-10-21
fix pcn-k8s docker build document

goldenrye:
- add extra line to make literal code block display appropriately
- fix the docker build parameter, pcn-k8s should be passed to docker as argument
  instead of environment


frisso:
@alacuku Don't you experience this issue while playing with K8s?

alacuku:
No, I don't. I always used a pre-made docker image of pcn-k8s.

PR #229, opened 2019-10-26
Adding Polycube extractor from docker image to host's fs

michelsciortino:
This script copies all the binaries, headers and libraries from a running Polycube docker image to the host's filesystem. This can be used in Jenkins on the Kubernetes slave nodes after pulling the testing Polycube image in order to correctly run the tests.

The script assumes the running image's name is 'polycubed'.
It is possible to specify a different name as an argument
e.g.: `$ polycube_from_docker_to_host.sh myPolycubeContainer`

### DIFF
--- a/scripts/polycube_from_docker_to_host.sh
+++ b/scripts/polycube_from_docker_to_host.sh
@@ -1,0 +1,51 @@
+#!/bin/bash
+
+#
+# This script copies all the binaries, headers and libraries from a running Polycube docker image to the host's filesystem.
+#
+# By default the script assumes the running container name is 'polycubed'.
+# It is possible to specify a different name as an argument
+# e.g.: $ polycube_from_docker_to_host.sh myPolycubeContainer
+#
+
+image_name="polycubed"
+
+if ! test -z "$1"
+then
+    image_name=$1
+fi
+
+# Copying binaries
+sudo docker cp $image_name:/usr/local/bin /usr/local
+
+# Copying headers
+sudo mkdir -p /usr/include/polycube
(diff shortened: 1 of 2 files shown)


scripts/polycube_from_docker_to_host.sh
```diff
@@ -0,0 +1,38 @@
...
+do
+    sudo docker cp $image_name:/usr/lib/x86_64-linux-gnu/$lib /usr/lib/x86_64-linux-gnu
+done
+
+for lib in $(sudo docker exec $image_name ls /lib/x86_64-linux-gnu/ | grep -P "libnl")
+do
+    sudo docker cp $image_name:/lib/x86_64-linux-gnu/$lib /lib/x86_64-linux-gnu
+done
+
+# Copying base yang models
```
frisso: Are we sure that we do need to copy YANG models?

scripts/polycube_from_docker_to_host.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/bash
```
frisso: Add a comment saying what is this script about and when will it be used.

tests/polycube_from_docker_to_host.sh
```diff
@@ -0,0 +1,46 @@
+#!/bin/bash
+
+#
+# This script copies all the binaries, headers and libraries from a running Polycube docker image to the host's filesystem.
```
frisso: Michel, 
1) the line is too long;
2) it is not enough to say what it does; it would be better to say also *why* this is needed.
Thanks!

PR #232, opened 2019-10-30
fix docker file issue 

goldenrye:
the docker image is built against the /tmp/polycube directory which is copied from the latest source code, but in the based image the /tmp/polycube is already existing. In the case the based image is not the latest code then the docker image doesn't contains the latest code. 

This PR make a clean sandbox before copying and building the docker image against the source code, otherwise the docker image could be built against the old sandbox left-over in the base image. Another way to do that is to clean the base image /tmp/polycube directory but anyway make a clean sandbox is always a good habit without any assumption.

### DIFF
--- a/Dockerfile_base_image
+++ b/Dockerfile_base_image
@@ -1,7 +1,7 @@
 # syntax = tonistiigi/dockerfile:runmount20180618
 FROM ubuntu:18.04
-RUN --mount=target=/polycube cp -r /polycube /tmp/polycube && \
-cd /tmp/polycube && \
-SUDO="" WORKDIR="/tmp/dev" \
+COPY scripts/pre-requirements.sh /tmp/
+RUN cd /tmp/ && \
 apt update && \
-./scripts/pre-requirements.sh
+SUDO="" WORKDIR="/tmp/dev" ./pre-requirements.sh && \
+rm -f /tmp/pre-requirements.sh
--- a/scripts/pre-requirements.sh
+++ b/scripts/pre-requirements.sh
@@ -5,7 +5,7 @@ then
 fi
 mkdir -p $WORKDIR
 
-$SUDO sudo apt update
+$SUDO apt update
 $SUDO bash -c "apt install --allow-unauthenticated -y wget gnupg2 software-properties-common"
 
 # golang v1.12 still not available in repos


frisso:
@michelsciortino Never got this problem?

michelsciortino:
@goldenrye is right.
I'm wondering why the source code is copied inside /tmp/polycube in the base image; the base image should just pre-install all the Polycube dependencies in a clean Ubuntu distro.

goldenrye:
Modified both the base image docker file -- don't contain the polycube source code

Dockerfile_base_image
```diff
@@ -1,7 +1,8 @@
 # syntax = tonistiigi/dockerfile:runmount20180618
 FROM ubuntu:18.04
-RUN --mount=target=/polycube cp -r /polycube /tmp/polycube && \
-cd /tmp/polycube && \
+RUN --mount=target=/polycube \
```
mauriciovasquezbernal: Maybe you can use `COPY scripts/pre-requirements.sh /pre-requirements.sh` instead of mounting the whole polycube folder?

Dockerfile
```diff
@@ -3,7 +3,7 @@ FROM polycubebot/base_image:latest
 ARG DEFAULT_MODE=default
 ENV MODE=$DEFAULT_MODE
 RUN echo "The mode is $MODE"
-RUN --mount=target=/polycube cp -r /polycube /tmp/polycube && \
+RUN --mount=target=/polycube rm -fr /tmp/polycube && cp -r /polycube /tmp/polycube && \
```
mauriciovasquezbernal: Is `rm -fr /tmp/polycube` really needed?

michelsciortino:
I think the `/tmp/pre-requirements.sh` file should be removed after having run it.

goldenrye:
> I think the `/tmp/pre-requirements.sh` file should be removed after having run it.

Done

frisso:
Jianwen, shouldn't we simply to "rebase", instead of doing "Merge branch 'master' into jpi-dockerfile"?

goldenrye:
> Jianwen, shouldn't we simply to "rebase", instead of doing "Merge branch 'master' into jpi-dockerfile"?

goldenrye:
> > Jianwen, shouldn't we simply to "rebase", instead of doing "Merge branch 'master' into jpi-dockerfile"?

Yes, I correct it.

PR #238, opened 2019-11-20
Fixed NAT service and modified utils function

DavideAG:
This pull request will fix NAT service.
We need to investigate if the problem persists for future transparent services.

### DIFF
--- a/src/services/pcn-nat/src/Nat_dp.c
+++ b/src/services/pcn-nat/src/Nat_dp.c
@@ -10,7 +10,7 @@ static int handle_rx(struct CTXTYPE *ctx, struct pkt_metadata *md) {
   void *data_end = (void *)(long)ctx->data_end;
 
   struct eth_hdr *eth = data;
-  if (data + sizeof(*eth) > data_end)
+  if ( (void *)eth + sizeof(*eth) > data_end )
     goto DROP;
 
   pcn_log(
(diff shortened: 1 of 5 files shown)


src/services/pcn-nat/src/Nat_dp.c
```diff
@@ -61,8 +61,9 @@ static int handle_rx(struct CTXTYPE *ctx, struct pkt_metadata *md) {
 
   switch (ip->protocol) {
   case IPPROTO_TCP: {
-    struct tcphdr *tcp = data + sizeof(*eth) + sizeof(*ip);
-    if (data + sizeof(*eth) + sizeof(*ip) + sizeof(*tcp) > data_end)
+    uint8_t header_len = 4 * ip->ihl;
+    struct tcphdr *tcp = data + sizeof(*eth) + header_len;
+    if (data + sizeof(*eth) + header_len + sizeof(*tcp) > data_end)
```
frisso: Would it be possible to write this as:
  if ( (tcp+ sizeof(*tcp)) > data_end)
The same for other similar lines below.

sebymiano: Definitely feasible, although the way it is written now is in line with the kernel code (e.g., https://github.com/torvalds/linux/blob/2027cabe6afea5d471401ec704b65ce18f958fdc/samples/bpf/parse_simple.c#L36)

However, if you decide to change it I would suggest to write it in this way:
```
if ( (void *)tcp + sizeof(*tcp) > data_end)
```
this would avoid annoying warnings on different pointer types comparisons.

DavideAG: Fine, thank you for your help @sebymiano. I will do it

frisso: @sebymiano thanks for the comment.
I was just wondering how smart is the compiler. Since "*tcp = data + sizeof(*eth) + header_len;" was computed the line before, I was simply proposing the reuse that result instead of computing it again in the next line.
So, if the compiler is smart, this code rewriting is useless. If it is not so smart, this saves a couple of ASM instructions (and, to me, looks more clear).

PR #239, opened 2019-11-22
Add standard output messages for successful tests

palexster:


palexster:
Do not merge, This introduces a bug in the script!

PR #249, opened 2019-12-05
fix polycubed crash & add a CLI to check the cubes given a interface

goldenrye:
When attach multiple cubes to one interface with "position=last", "before=xxx", "after=xxx", system crashes. This PR fixes the crashing issue and add a CLI to check the cubes attaches to a interface given a name.

### DIFF
--- a/src/polycubed/src/extiface.cpp
+++ b/src/polycubed/src/extiface.cpp
@@ -180,39 +180,82 @@ void ExtIface::set_next(uint16_t next, ProgramType type) {
   }
 }
 
+TransparentCube* ExtIface::get_next_cube(ProgramType type) {
+  auto next = get_next(type);
+  for (auto &it : cubes_) {
+      if (it->get_index(type) == next) {
+          return it;
+      }
+  }
+  return NULL;
+}
+
+uint16_t ExtIface::get_next(ProgramType type) {
+  int zero = 0;
+  unsigned int value = 0;
+  auto program = (type == ProgramType::INGRESS) ? &ingress_program_ : &egress_program_;
+  auto index_map = program->get_array_table<uint32_t>("index_map");
+  auto st = index_map.get_value(zero, value);
+  if (st.code() == -1) {
+    logger->error("failed to get interface {0} nexthop", get_iface_name());
+  }
+
+  return value & 0xffff;
+}
+
+std::vector<std::string> ExtIface::get_service_chain(ProgramType type) {
+  std::vector<std::string> chain;
+  TransparentCube *cube;
+  auto nh = get_next(type);
+
+  while (nh != 0xffff) {
+      cube = NULL;
+      for (auto c : cubes_) {
+          if (c->get_index(type) == nh) {
+              cube = c;
+              break;
+          }
+      }
+      chain.push_back(cube ? cube->get_name() : "unknown");
+      nh = cube->get_next(type);
+  }
+  chain.push_back(nh==0xffff ? "stack" : "unknown");
+
+  return chain;
+}
+
 void ExtIface::update_indexes() {
   int i;
 
   // TODO: could we avoid to recalculate in case there is not peer?
 
-  std::vector<uint16_t> ingress_indexes(cubes_.size());
-  std::vector<uint16_t> egress_indexes(cubes_.size());
+  std::vector<uint16_t> ingress_indexes(cubes_.size(), 0xffff) ;
+  std::vector<uint16_t> egress_indexes(cubes_.size(), 0xffff);
 
   for (i = 0; i < cubes_.size(); i++) {
     ingress_indexes[i] = cubes_[i]->get_index(ProgramType::INGRESS);
     egress_indexes[i] = cubes_[i]->get_index(ProgramType::EGRESS);
   }
 
-  // ingress chain: NIC -> cube[N-1] -> ... -> cube[0] -> stack (or peer)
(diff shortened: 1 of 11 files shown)


goldenrye:
> That's a very good point, Mauricio.
> Given also the comment made by @palexster , would it be possible, for you that know how it should work, define a test that can be used to make sure this behaves as expected? Thanks!

goldenrye:
Add some code to traverse and dump the interface's service chain, this will help to verify that the cubes attached to the interface was linked in a appropriate order. Here is an example:
# polycubectl firewall add fw1
# polycubectl ddosmitigator dm1
# polycubectl attach fw1 ens9
# polycubectl attach dm1 ens9 position=first

after attaching a firewall cube and a ddosmitigator cube to a interface "ens9", we can verify the order via the command
# polycubectl topology show ens9
name: ens9
ingress service chain: 1 ---> 3 ---> stack
egress service chain: 2 ---> 0 ---> stack

cubes, ingress_index, egress_index:
 dm1,   3,      0
 fw1,    1,      2

From the output we know the packet will go through firewall and then ddosmitigator.

I will add a test script to verify different scenarios.

mauriciovasquezbernal:
@frisso Sorry for not being completely clear, I meant the lack of infrastructure for unit test. I know that there is Jenkins and also that there are a lot of end-2-end tests, but this one would need some imagination to be tested with the current infrastructure.

What we need to do is to attach multiple transparent cubes to an interface/port and send traffic to that interface to know which order are the cubes crossed. We can test if dropping is working, if sending to the stack is working, but how do you test that the cubes are called in a given order? I tried multiple times to create a test but I didn't find a solution. I was thinking that connecting multiple transparent-helloworld instances and then gathering their terminal output could help.

mauriciovasquezbernal:
> Add some code to traverse and dump the interface's service chain, this will help to verify that the cubes attached to the interface was linked in a appropriate order. Since the topology(service chain) is a real reflection of the pipeline in the data plane data, no real traffic is needed to test this topology functionality. Here is an example:
> 
> > polycubectl firewall add fw1
> > polycubectl ddosmitigator dm1
> > polycubectl attach fw1 ens9
> > polycubectl attach dm1 ens9 position=first

I think it is worth to clarify what the first word means here. This support was designed considering that `first` is the cube that is called **first** in **incoming** traffic. So in this case dm1 would be called first and then fw1 for ingress traffic. Egress traffic is the other way around (although in this particular example it is not relevant because those services don't implement an egress hook).
> 
> after attaching a firewall cube and a ddosmitigator cube to a interface "ens9", we can verify the order via the command
> 
> > polycubectl topology show ens9
> > name: ens9
> > ingress service chain: 1 ---> 3 ---> stack
> > egress service chain: 2 ---> 0 ---> stack
> > cubes, ingress_index, egress_index:
> > dm1,   3,      0
> > fw1,    1,      2
> 
> From the output we know the packet will go through firewall and then ddosmitigator.
> 

According to that, both cubes are called in the same order for ingress and egress traffic (first fw1 and then dm1). Is that ok?
Also, considering the clarification I gave above about `first` meaning, this one is inverted, right?

> As mentioned we don't need to send real traffic to test this topology functionality, I will add a few test scripts to verify different scenarios.

I'd suggest that you send real traffic to check the order, at the end this is the only source of truth that would confirm that cubes are called in the correct order.

goldenrye:
I don't know why replying to a particular comment doesn't work in my system, so i answer some the questions here:

1. regarding the ingress and egress order should be in reverse I have fixed the issue and merge it to 2nd commit;
2. add a test case for testing this feature and test it manually locally, it works as expected,
jpi@polycube-dev:/tmp$ ./test.sh
+ set -e
+ trap clean EXIT
+ polycubectl transparenthelloworld add th1 loglevel=TRACE
+ polycubectl transparenthelloworld add th2 loglevel=TRACE
+ polycubectl transparenthelloworld add th3 loglevel=TRACE
+ polycubectl transparenthelloworld add th4 loglevel=TRACE
+ polycubectl transparenthelloworld add th5 loglevel=TRACE
+ polycubectl attach th3 ens9
+ polycubectl attach th1 ens9 position=first
+ polycubectl attach th5 ens9 position=last
+ polycubectl attach th2 ens9 after=th1
+ polycubectl attach th4 ens9 before=th5
+ polycubectl topology show ens9
name: ens9
ingress service chain: 1 ---> 3 ---> 6 ---> 8 ---> 10 ---> stack
egress service chain: 11 ---> 9 ---> 7 ---> 5 ---> 2 ---> stack

cubes, ingress_index, egress_index:

 th1,   1,      2
 th2,   3,      5
 th3,   6,      7
 th4,   8,      9
 th5,   10,     11

+ polycubectl del th1
+ polycubectl del th2
+ polycubectl del th3
+ polycubectl del th4
+ polycubectl del th5
+ set +x
SUCCESS

frisso:
> @frisso Sorry for not being completely clear, I meant the lack of infrastructure for unit test. I know that there is Jenkins and also that there are a lot of end-2-end tests, but this one would need some imagination to be tested with the current infrastructure.
> 
> What we need to do is to attach multiple transparent cubes to an interface/port and send traffic to that interface to know which order are the cubes crossed. We can test if dropping is working, if sending to the stack is working, but how do you test that the cubes are called in a given order? I tried multiple times to create a test but I didn't find a solution. I was thinking that connecting multiple transparent-helloworld instances and then gathering their terminal output could help.

Got your point now.
Yes, we're still missing unit testing.

PR #250, opened 2019-12-07
install appropriate services for different products

goldenrye:
Explicitly turn off the simpleforwarder, transparenthelloworld and synflood service in install script so these services won't be installed in pcn-k8s and iptables release.


frisso:
@goldenrye Shouldn't DENABLE_SERVICE_DDOSMITIGATOR=ON in Kubernetes mode? We're not using it right now, but it may be useful. The same for packet capture.

goldenrye:
> Shouldn't DENABLE_SERVICE_DDOSMITIGATOR=ON in Kubernetes mode? We're not using it right now, but it may be useful. The same for packet capture.

@frisso cannot agree with that more, and modified the configuration.

frisso:
👍 
@goldenrye feel free to merge. Thanks!

Issue #233, opened 2019-11-08
[BUG] Failing test in pcn-iptables service


michelsciortino:
## Describe the bug
The `local_test_interfaces1.sh` has been renamed as `flaky_local_test_interfaces1.sh` in order to be excluded from the `pcn-iptables` tests execution as it fails on the new Jenkins slaves.

The script tests the connectivity between two namespaces by using the `ping` command, expecting to have packet loss when the forwarding policy is not set to `ACCEPT` on both `veth1` and `veth2` interfaces exposed by the two namespaces.

When the forwarding policy is set to `DROP` on both the interfaces, the echo-requests sent from one namespace to the other are correctly dropped.

When the forwarding policy is set to `ACCEPT` on one of the two interfaces by running
`pcn-iptables -A FORWARD -i veth1 -j ACCEPT`
or
`pcn-iptables -A FORWARD -i veth2 -j ACCEPT`
the echo-requests receive responses instead of having packet loss in the forwarding.

## To Reproduce
Install required tools:
```
sudo apt-get install git net-tools
```

Clone the Polycube repository:
```
git clone https://github.com/polycube-network/polycube
cd polycube
git submodule update --init --recursive
```
Launch the automatic install script:
```
./scripts/install.sh
```
Execute the failing test:
```
cd src/services/pcn-iptables/test
./flaky_local_test_interfaces1.sh 
```
Show test result:
```
echo $? # will display 1
```
## Expected behavior
The test executes
```
pcn-iptables -P FORWARD DROP
pcn-iptables -A FORWARD -i veth1 -j ACCEPT
test_fail sudo ip netns exec ns1 ping 10.0.2.1 -c 2 -W 2
```
expecting the packets to be dropped as forwarding is enabled only on interface `veth1` and the echo requests should not receive any reply.

## Please tell us about your environment: 
1. OS details:
```
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=18.04
DISTRIB_CODENAME=bionic
DISTRIB_DESCRIPTION="Ubuntu 18.04.2 LTS"
NAME="Ubuntu"
VERSION="18.04.2 LTS (Bionic Beaver)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 18.04.2 LTS"
VERSION_ID="18.04"
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
[...]

PR #252, opened 2019-12-08
fix pcn-iptables interface matching issue

goldenrye:
pcn-iptables uses 0 as wild card in port's hash map, which conflicts if
a port was assigned with id 0. Don't assign port id 0 to any port should
be able to fix the issue #233 .


src/polycubed/src/cube.cpp
```diff
@@ -39,8 +39,10 @@ Cube::Cube(const std::string &name, const std::string &service_name,
   forward_chain_ = std::unique_ptr<ebpf::BPFArrayTable<uint32_t>>(
       new ebpf::BPFArrayTable<uint32_t>(forward_));
 
-  // add free ports
-  for (uint16_t i = 0; i < _POLYCUBE_MAX_PORTS; i++)
+  /* add free ports, port id 0 is reserved by pcn-iptables as wild card, so
+   * it won't be put into the free pool
+   */
+  for (uint16_t i = 1; i < _POLYCUBE_MAX_PORTS+1; i++)
```
frisso: @mauriciovasquezbernal Are we safe here? We're cycling till _POLYCUBE_MAX_PORTS+1, while internal variables (e.g., BPF_TABLE_SHARED("array", int, u32, forward_chain_, _POLYCUBE_MAX_PORTS); ) are dimensioned with _POLYCUBE_MAX_PORTS.

mauriciovasquezbernal: It's not safe. If a cube gets the `_POLYCUBE_MAX_PORTS` port id it'll be **not** able to use that port because the lookup in that map will fail.

goldenrye:
Ok thanks for all the comments and suggestions. I have modify the wild card from 0 to 0xffff and add some static assert to ensure no conflicting with regular port.

BTW I think we should decouple the port id with the value _POLYCUBE_MAX_PORT, there is no reason to assume the port id is less than _POLYCUBE_MAX_PORT - I didn't actually see any code depends on this assumption.Ok thanks for all the comments and suggestions. I have modify the wild card from 0 to 0xffff and add some static assert to ensure no conflicting with regular port.

BTW I think we should decouple the port id with the value _POLYCUBE_MAX_PORT, there is no reason to assume the port id is less than _POLYCUBE_MAX_PORT - I didn't actually see any code depends on this assumption.

mauriciovasquezbernal:
> BTW I think we should decouple the port id with the value _POLYCUBE_MAX_PORT, there is no reason to assume the port id is less than _POLYCUBE_MAX_PORT - I didn't actually see any code depends on this assumption.

Each cube has an array map that saves the relationship `port_id` -> `endpoint (cube_id | port_id)`:

https://github.com/polycube-network/polycube/blob/d30346200f26ee88a6f89469f64abab8cba48693/src/polycubed/src/cube.cpp#L309-L310

The result of the lookup in that table is a pair of values, the destination cube and port of the packet, based on that a tail call is done to the next cube:

https://github.com/polycube-network/polycube/blob/ac2e21d693df8724f33702b105c125183c61b293/src/polycubed/src/cube_tc.cpp#L301

In order to remove the requirement that `port_id` < `_POLYCUBE_MAX_PORT` that map would have to be changed to a hash map.  I think it is not worth to do because introduces overhead and cannot see what is the benefit of removing that constraint.

Could you please clarify why you think this constraint should be removed?

goldenrye:
> In order to remove the requirement that `port_id` < `_POLYCUBE_MAX_PORT` that map would have to be changed to a hash map. I think it is not worth to do because introduces overhead and cannot see what is the benefit of removing that constraint.
> 
> Could you please clarify why you think this constraint should be removed?

If the forwarding table use port id as the index, the modification is not worthy, I don't have very strong case to support the modification. My argument is based on that port id is only used as value and any user referring to the value shouldn't have any assumption the range of the value.

mauriciovasquezbernal:
> My argument is based on that port id is only used as value and any user referring to the value shouldn't have any assumption the range of the value.

I don't know how pcn-iptables works but in the general cube idea the `port_id` is an internal reference and should be never exposed to the user. Cubes' ports are identified by the port name.

goldenrye:
@mauriciovasquezbernal @sebymiano all the tests are passed, please review the code, thanks!

Issue #184, opened 2019-07-22
[BUG] Firewall creation fails if including more than fw's name in the JSON 


SunSince90:
## Describe the bug
Depending on what is included, firewall creation may fail or Polycube may crash.

These two JSON are fine:
`{"name":"fw-test","interactive":false}`
`{"name":"fw-test"}`

Creating a firewall and specifying the default action of the chains will fail (Ingress chain does not exist):
`{"name":"fw-test","chain":[{"name":"ingress","default":"forward"},{"name":"egress","default":"forward"}]}`

Polycube crashes entirely if accept-established is set to ON:
`{"name":"fw-test","accept-established":"ON"}`

## To Reproduce
First two cases:
`curl -d '{"name":"fw-test"}' -H "Content-Type: application/json" -X POST http://127.0.0.1:9000/polycube/v1/firewall/fw-test/`
`curl -d '{"name":"fw-test","interactive":false}' -H "Content-Type: application/json" -X POST http://127.0.0.1:9000/polycube/v1/firewall/fw-test/`

Including the chains:
`curl -d '{"name":"fw-test","chain":[{"name":"ingress","default":"forward"},{"name":"egress","default":"forward"}]}' -H "Content-Type: application/json" -X POST http://127.0.0.1:9000/polycube/v1/firewall/fw-test/`

Accept-established set to ON:
`curl -d '{"name":"fw-test","accept-established":"ON"}' -H "Content-Type: application/json" -X POST http://127.0.0.1:9000/polycube/v1/firewall/fw-test/`

## Expected behavior
`201` was expected for all cases.

## Please tell us about your environment: 
1. OS details: 
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=18.04
DISTRIB_CODENAME=bionic
DISTRIB_DESCRIPTION="Ubuntu 18.04.2 LTS"
NAME="Ubuntu"
VERSION="18.04.2 LTS (Bionic Beaver)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 18.04.2 LTS"
VERSION_ID="18.04"
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
VERSION_CODENAME=bionic
[...]

PR #256, opened 2019-12-15
fix firewall creation fails issue 

goldenrye:
Adjust the init order when creating firewall cube object so the parameter passed
by user can be populated, the will also avoid the crash due to object reference
before object initialization.

issue #184

### DIFF
--- a/src/services/pcn-firewall/src/Firewall.cpp
+++ b/src/services/pcn-firewall/src/Firewall.cpp
@@ -22,7 +22,6 @@ Firewall::Firewall(const std::string name, const FirewallJsonObject &conf)
   logger()->set_pattern("[%Y-%m-%d %H:%M:%S.%e] [Firewall] [%n] [%l] %v");
   logger()->info("Creating Firewall instance");
 
-  update(conf);
 
   /*Creating default INGRESS and EGRESS chains.*/
   ChainJsonObject chain;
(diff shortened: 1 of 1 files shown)


frisso:
Jianwen, just a question: any idea why this was never noticed in our tests?

goldenrye:
it happens in the Jenkins server,
http://130.192.225.104:8080/job/polycube-test-clean/job/PR-256/3/console

Maybe this is because the libyang dependencies were changed, I tried
locally the same issue will happen as well.

goldenrye:
Sorry just ignore my reply. I thought Professor is asking the libyang
dependency issue.

goldenrye:
> Jianwen, just a question: any idea why this was never noticed in our tests?

That's because we don't have test scripts to cover those cases.

If you execute the following command with the latest master branch build system will crash,
polycubectl firewall add fw-test accept-established=on

For default chain action, polycubectl doesn't allow to set the default chain action, so you won't be able run into this issue via CLI. But using curl command you should be able to reproduce the issue:
curl -d '{"name":"fw-test","chain":[{"name":"ingress","default":"forward"},{"name":"egress","default":"forward"}]}' -H "Content-Type: application/json" -X POST http://127.0.0.1:9000/polycube/v1/firewall/fw-test/

PR #259, opened 2020-01-15
A new elegant whiptail based documentation script for polycube

DavideAG:
This bash based script allows to show services documentation in a whiptail dialog.

launcher
service

### DIFF
--- a/scripts/man.sh
+++ b/scripts/man.sh
@@ -1,0 +1,34 @@
+#!/bin/bash
+shopt -s -o nounset
+
+count=0
+W=()
+_script="$(readlink -f ${BASH_SOURCE[0]})"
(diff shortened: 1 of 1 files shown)


frisso:
Thanks, Davide!
One minor comment: folder "Documentation" is for docs that are being pushed on readthedocs. So, users do not expect to find a script there, hence may not look for it and your work may go unnoticed.
I would suggest to commit this into the /script folder.

DavideAG:
> Thanks, Davide!
> One minor comment: folder "Documentation" is for docs that are being pushed on readthedocs. So, users do not expect to find a script there, hence may not look for it and your work may go unnoticed.
> I would suggest to commit this into the /script folder.

Fine, I will do it. Maybe I have to change a little piece of code in the script to select the right folder and support this

PR #267, opened 2020-02-07
Implementing maxPayload Pistache option personalization

michelsciortino:
Currently the Pistache `maxPayload`  option has not been managed in the initialization of the Pistache server, so the HTTP requests with a payload size bigger than the default Pistache' setting produce a `Request exceeded maximum buffer size` response error message, with strange behaviors on the subsequent HTTP requests.

This PR adds the `maxPayload` option in the initialization of the Pistache server and set it to a default value of [`64KB`](https://github.com/polycube-network/polycube/blob/473f505c4f252e98473ea97a83bf9a17fd5166e1/src/polycubed/src/rest_server.h#L56).

The value of `maxPayload` is then customized in the rest server setup with a value of [`1MB`](https://github.com/polycube-network/polycube/blob/473f505c4f252e98473ea97a83bf9a17fd5166e1/src/polycubed/src/polycubed.cpp#L277).

### DIFF
--- a/src/polycubed/src/rest_server.h
+++ b/src/polycubed/src/rest_server.h
@@ -52,7 +52,9 @@ class RestServer {
 
   static const std::string base;
 
-  void init(size_t thr = 1, const std::string &server_cert = "",
+  void init(size_t thr = 1,
+            size_t max_payload_size = 64*1024,
+            const std::string &server_cert = "",
             const std::string &server_key = "",
             const std::string &root_ca_cert = "",
             const std::string &white_cert_list_path = "",
--- a/src/polycubed/src/polycubed.cpp
+++ b/src/polycubed/src/polycubed.cpp
@@ -274,6 +274,7 @@ int main(int argc, char *argv[]) {
 
   // setup rest server
   int thr = 4;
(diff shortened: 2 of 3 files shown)


src/polycubed/src/rest_server.h
```diff
@@ -52,7 +52,9 @@ class RestServer {
 
   static const std::string base;
 
-  void init(size_t thr = 1, const std::string &server_cert = "",
+  void init(size_t thr = 1,
+            size_t max_payload_size = 64*1024,
```
frisso: Can you please add a comment that says how big is the payload? Just because you told me "I set the max payload to 1MB", but from the code it looks like 64KB, hence it looks confusing to me.

michelsciortino: 64KB is the default init value in the rest_server.h header;
The 1MB setting is in [polycubed.cpp](https://github.com/polycube-network/polycube/blob/473f505c4f252e98473ea97a83bf9a17fd5166e1/src/polycubed/src/polycubed.cpp#L277), where the rest server is being initialized with custom settings

PR #270, opened 2020-02-13
This code fixes two bugs about "packetcapture" cube:

guerol:
- activation of "packetcapture" on an interface makes it unreachable
- duplication of traffic in egress of an interface attached to
"packetcapture" service

On branch unreachable_interface_packetcapture_fixing
Changes to be committed:
	modified:   src/services/pcn-packetcapture/src/Packetcapture.cpp
	modified:   src/services/pcn-packetcapture/src/Packetcapture_dp_egress.c
	modified:   src/services/pcn-packetcapture/src/Packetcapture_dp_ingress.c

### DIFF
--- a/src/services/pcn-packetcapture/src/Packetcapture.cpp
+++ b/src/services/pcn-packetcapture/src/Packetcapture.cpp
@@ -216,7 +216,7 @@ void Packetcapture::packet_in(polycube::service::Direction direction,
     }
     break;
   }
-  send_packet_out(pkt, direction, false);
+  //send_packet_out(pkt, direction, false);
 }
 
 PacketcaptureCaptureEnum Packetcapture::getCapture() {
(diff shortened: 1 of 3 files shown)


DavideAG:
As reported by @sebymiano we need to find a solution to forward the packet to the slow path in order to allow it to work both if it is attached to a pysical netdevice or a virtual netdev.
I had found an abnormal API behavior that was introduced by this PR.

However, I will try to think of a (possibly transparent) solution for this problem.
@sebymiano maybe we could talk about it

goldenrye:
Can somebody describe the issue more clearly so the issue and the solution can be tracked?

frisso:
This PR is very old and it didn't solve the issue in all the conditions.
Solution has been provided by #292 : thanks, @FedeParola .
So, closing this issue.

PR #273, opened 2020-02-17
Filter for packetcapture service

ricjames11:
the filter is tcp dump like (especially the syntax to insert it). In particular from the string of the filter inserted by the user, it will generate the eBPF code which will then be translated into C code to be injected into the dp file


src/services/pcn-packetcapture/src/cbpfTOc.cpp
```diff
@@ -0,0 +1,490 @@
+#include <iostream>
```
frisso: This file is the core of your work.
Please:
  (a) add some reference (link) to Cloudfare, where you get some ideas (people get upset when you don't give them credits)
  (b) add some comments. It may not be trivial to understand what's going on in this file.

src/services/pcn-packetcapture/src/cbpfTOc.cpp
```diff
@@ -0,0 +1,490 @@
...
+#include "cbpfTOc.h"
+#include "linux/bpf.h"
+#include <arpa/inet.h>
+#include <linux/filter.h>
+#include <map>
+#include <cstring>
```
frisso: Somewhere in your code (likely in the control plane, where you get the C equivalent of your filter), please add also a debug possibility where the C code is dumped. This helps debugging this service in case something goes wrong with your compiler.

Documentation/services/pcn-packetcapture/packetcapture.rst
```diff
@@ -1,14 +1,14 @@
 Packetcapture service
 =====================
```
frisso: It would be helpful to add another rst file with the overview of the implementation, i.e., how you can leverage libpcap syntax for filtering packets in eBPF.

frisso:
Riccardo, very good work.
I would suggest to align with the latest master (there has been some changes made recently by Davide here), then take care of the comments. Overall, looks good.
Thanks!

frisso:
@DavideAG Feel free to add your review.

Documentation/services/pcn-packetcapture/packetcapture.rst
```diff
@@ -56,61 +56,54 @@ Now the packetcapture service is attached to the port *toveth1* of the bridge *b
...
-- destination prefix
-- source port
-- destination port
-- later 4 protocol
+    polycubectl <service name> set filter=<string value>
+
+-if the filter contains only one word you can put it normally
+-if the filter contains more than a word you have to put the string inside the ""
+
+-if you want to capture all the traffic you can put as filter: all
```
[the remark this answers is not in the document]
ricjames11: the default filter is "RETURN RX_OK;"
Now it is specified in the documentation

src/services/pcn-packetcapture/datamodel/packetcapture.yang
```diff
@@ -46,37 +46,15 @@ module packetcapture {
...
-    }    
+  leaf snaplen {
+    type uint32;
+    default 262144;
+    description "Snapshot length";
+  }
+
+  leaf filter {
+    type string;
+    description "custom filter inserted at run-time";
```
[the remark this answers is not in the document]
ricjames11: I've changed it but I've used single quotes for host 1.2.3.4... otherwise error occurred

src/services/pcn-packetcapture/src/FindPCAP.cmake
```diff
@@ -0,0 +1,74 @@
+# - Try to find libpcap include dirs and libraries
```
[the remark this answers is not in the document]
ricjames11: Yes, I've removed it and also the line in the CMakeList and I added in the documentation in "How To Use" part the command to  install libpcap-dev

src/services/pcn-packetcapture/src/Packetcapture.cpp
```diff
@@ -435,4 +397,63 @@ void Packetcapture::to_timeval(std::chrono::system_clock::duration& d, struct ti
...
+    return this->filter;
+}
+
+uint32_t Packetcapture::getSnaplen() {
+    return this->snaplen;
+}
+
+void Packetcapture::setSnaplen(const uint32_t &value) {
+    this->snaplen = value;
+    if (!bootstrap)
```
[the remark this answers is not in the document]
ricjames11: It was a typo of Davide's code, now it is no more necessary

src/services/pcn-packetcapture/src/Packetcapture.h
```diff
@@ -45,7 +44,15 @@ class Packetcapture : public PacketcaptureBase {
  std::string random_number;
  std::chrono::system_clock::time_point timeP;
  std::chrono::nanoseconds temp_offset;
+ bool bootstrap = true;
+ std::string filter;
+ uint32_t snaplen = 262144;  /* 65535 for no sliced packets */
```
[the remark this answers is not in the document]
ricjames11: The only places where this number appears is in the Packetcapture.h and in the yang model

Documentation/services/pcn-packetcapture/packetcapture.rst
ricjames11: Something like this one https://blog.cloudflare.com/xdpcap/ ?
or maybe just a reference to the link above in the actual documentation?

.gitmodules
```diff
@@ -7,3 +7,15 @@
 	path = src/components/iptables/iptables
 	url = https://github.com/polycube-network/iptables.git
 	branch = polycube
+[submodule "polycube-codegen"]
```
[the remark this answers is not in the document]
ricjames11: No, sorry it is not necessary, I deleted it

Documentation/services/pcn-packetcapture/packetcapture.rst
```diff
@@ -56,61 +56,54 @@ Now the packetcapture service is attached to the port *toveth1* of the bridge *b
...
     # Example of the layer 4 protocol filter
-    polycubectl mysniffer filters set l4proto=tcp
+    polycubectl mysniffer set filter=tcp
 
 Snaplen filter
 ^^^^^^^^^^^^^^
 ::
     
     # Example of the snaplen filter
     #  In this case we capture only the first 80 bytes of each packet
```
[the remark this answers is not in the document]
ricjames11: Here is the code for the tcp filter:
```
L0:	//(ldh)
	if((data + 14) > data_end){
		return RX_DROP;
	}
	a = ntohs(* ((uint16_t *) &data[12]));
L1:	if(a == 0x086dd) {
		goto L2;
	}else{
		goto L7;
	}
L2:	//(ldb)
	if((data + 21) > data_end){
		return RX_DROP;
	}
	a = * ((uint8_t *) &data[20]);
L3:	if(a == 0x06) {
		goto L10;
	}else{
		goto L4;
	}
L4:	if(a == 0x02c) {
		goto L5;
	}else{
		goto L11;
	}
L5:	//(ldb)
	if((data + 55) > data_end){
		return RX_DROP;
	}
	a = * ((uint8_t *) &data[54]);
L6:	if(a == 0x06) {
		goto L10;
	}else{
		goto L11;
	}
L7:	if(a == 0x0800) {
		goto L8;
	}else{
		goto L11;
	}
L8:	//(ldb)
	if((data + 24) > data_end){
		return RX_DROP;
	}
	a = * ((uint8_t *) &data[23]);
L9:	if(a == 0x06) {
		goto L10;
	}else{
		goto L11;
	}
L10:	return pcn_pkt_controller(ctx, md, reason);
L11:	return RX_OK;
```

ricjames11:
I've done some fix according to the comments, are still missing:
- rst file for filter documentation
- debug possibility of the filter C code

DavideAG:
> @DavideAG Feel free to add your review.

Thank you, I'm doing it

src/services/pcn-packetcapture/src/Packetcapture_dp_ingress.c
```diff
@@ -70,128 +53,15 @@ BPF_ARRAY(packet_timestamp, uint64_t, 1);
  */
 BPF_ARRAY(working, uint8_t, 1);
```
DavideAG: where are you using this BPF_ARRAY?
Are you able to differentiate the ebpf code that is activated? For example, can we capture only outgoing traffic and not incoming traffic?

src/services/pcn-packetcapture/src/Packetcapture_dp_egress.c
```diff
@@ -70,128 +53,15 @@ BPF_ARRAY(packet_timestamp, uint64_t, 1);
  */
 BPF_ARRAY(working, uint8_t, 1);
```
DavideAG: Same here of Packetcapture_dp_ingress.c

src/services/pcn-packetcapture/src/Packetcapture.cpp
```diff
@@ -165,23 +173,22 @@ void Packetcapture::writeDump(const std::vector<uint8_t> &packet){
...
+    p->setCapturelen(getSnaplen());
   } else {
     p->setCapturelen(p->getPacketlen());
   }
   p->setRawPacketData(packet);
 
   struct pcap_pkthdr *pkt_hdr = new struct pcap_pkthdr;
-  pkt_hdr->ts_sec = p->getTimestampSeconds();
-  pkt_hdr->ts_usec = p->getTimestampMicroseconds();
+  pkt_hdr->ts = tp;
```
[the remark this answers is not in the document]
ricjames11: Hi @DavideAG, thanks for the review

My problem here was that in the library pcap.h (that i need for pcap function for the filter) is defined the same struct but with other variable, in particolar here is the def. in pcap.h:
```
struct pcap_pkthdr {
	struct timeval ts;	/* time stamp */
	bpf_u_int32 caplen;	/* length of portion present */
	bpf_u_int32 len;	/* length this packet (off wire) */
};
```
here yours:
```
struct pcap_pkthdr {
	bpf_u_int32 ts_sec;	
        bpf_u_int32 ts_usec;	
	bpf_u_int32 caplen;	/* length of portion present */
	bpf_u_int32 len;	/* length this packet (off wire) */
};
```
but struct timeval is defined as:
```
struct timeval {
  __time_t tv_sec;		/* Seconds.  */
  __suseconds_t tv_usec;	/* Microseconds.  */
};
```
so I think is the same,  I only joined the two things
everything else is the same as your code

src/services/pcn-packetcapture/src/Packetcapture_dp_ingress.c
ricjames11: I can capture only outgoing or incoming or both because I inject the filterCode according to capture direction inserted by the user, if I capture only in one way in the other way the filter is "RETURN RX_OK;"

if (CapStatus == 3 || CapStatus == 2) {
    reload(codeINGRESS,0,ProgramType::INGRESS);
    reload(codeEGRESS,0,ProgramType::EGRESS);
  } else {
      if (CapStatus == 0) {
        reload(codeINGRESS,0,ProgramType::INGRESS);
      } else {
          if (CapStatus == 1) {
            reload(codeEGRESS,0,ProgramType::EGRESS);
          }
      }
  }

DavideAG: Well, you should then delete all references to using "working" maps if you no longer use them. Look for example in the fast paths of the service (ingress and egress) and in packetcapture.cpp where they are updated. you should also delete the definition of the associated enum

src/services/pcn-packetcapture/src/Packetcapture.cpp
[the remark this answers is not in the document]
DavideAG: I see your problem.
I tried your solution too, but I had problems when I was dumping packages directly into the server's file system (so without using the packetcapture client written in python).

Were you able to read correctly with wireshark a dump written directly from packetcapture? It should be stored in the tmp folder.
In my case, using the Linux struct timeval, the dumps produced by the packetcapture in the server file system were not readable.

I finally found a solution regarding the correct way to dump packages into the pcap file.
The problem is related to the byte size of the struct timeval variables.
These must be forced to 32bit as suggested by wireshark wiki.
https://wiki.wireshark.org/Development/LibpcapFileFormat#Record_.28Packet.29_Header


However, regardless of the structure you use, **the packet timestamp must be acquired in the fast path** of the service and **not in the slow path** otherwise it will not be accurate.

src/services/pcn-packetcapture/src/cbpfTOc.cpp
ricjames11: I've added `logger()->debug("FilterCode:\n{0}", filterCode);`

src/services/pcn-packetcapture/src/Packetcapture_dp_ingress.c
ricjames11: Ok perfect done, thanks

src/services/pcn-packetcapture/src/Packetcapture.cpp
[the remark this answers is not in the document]
ricjames11: No I'm not able, when I try to open the dump file in tmp folder wireshark say:

> The capture file appears to be damaged or corrupt.
> (pcap: File has 639023-byte packet, bigger than maximum of 262144)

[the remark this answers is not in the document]
ricjames11: I've changed the code and now I take the packet timestamp in the fast path as you also done

src/services/pcn-packetcapture/src/Packetcapture.cpp
```diff
@@ -17,56 +17,36 @@
...
-};
-
 Packetcapture::Packetcapture(const std::string name, const PacketcaptureJsonObject &conf)
   : TransparentCube(conf.getBase(), { packetcapture_code_ingress }, { packetcapture_code_egress }),
     PacketcaptureBase(name) {
-  
+
   /*
    * The timestamp of the packet is acquired in the fast path using 'bpf_ktime_get_ns()'
    * This function returns the time elapsed since system booted, in nanoseconds.
```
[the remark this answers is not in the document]
ricjames11: No change needed with the new implementation

ricjames11:
I've updated the code with new fix:
- I've added a new documentation for describing the filter procedure
- I've added a debug logger for print the C code of the filter
- I've fixed the problem with the timestamp as suggested by Davide, and now I take the timestamp in the fast path
- I've deleted some not needed structure
- I've corrected an error that arised the first time service start, so I moved the default filter injection in the updatefilter() function and now everything work right

src/services/pcn-packetcapture/src/Packetcapture.cpp
```diff
@@ -363,54 +331,27 @@ void Packetcapture::attach() {
...
-  auto ft_fast_in = get_array_table<filters_table>("filters_map", 0, ProgramType::INGRESS);
-  auto ft_fast_out = get_array_table<filters_table>("filters_map", 0, ProgramType::EGRESS);
+    if (CapStatus == 3 || CapStatus == 2) {
+      reload(codeINGRESS,0,ProgramType::INGRESS);
+      reload(codeEGRESS,0,ProgramType::EGRESS);
+    } else {
+        if (CapStatus == 0) {
+          reload(codeINGRESS,0,ProgramType::INGRESS);
+        } else {
+            if (CapStatus == 1) {
```
DavideAG: maybe you don't need this if statement

src/services/pcn-packetcapture/src/Packetcapture.cpp
```diff
@@ -435,4 +376,65 @@ void Packetcapture::to_timeval(std::chrono::system_clock::duration& d, struct ti
...
+  return this->filter;
+}
+
+uint32_t Packetcapture::getSnaplen() {
+  return this->snaplen;
+}
+
+void Packetcapture::setSnaplen(const uint32_t &value) {
+    this->snaplen = value;
+    updateFilter();
```
DavideAG: Why you need to update the Filter here?

src/services/pcn-packetcapture/src/Packetcapture.cpp
```diff
@@ -435,4 +376,65 @@ void Packetcapture::to_timeval(std::chrono::system_clock::duration& d, struct ti
...
+    }
+    updateFilter();
+}
+
+std::string Packetcapture::getFilter() {
+  return this->filter;
+}
+
+uint32_t Packetcapture::getSnaplen() {
+  return this->snaplen;
```
DavideAG: why did you introduce an additional snaplen value and not use the globalheader class?

src/services/pcn-packetcapture/src/Packetcapture.h
```diff
@@ -45,7 +44,15 @@ class Packetcapture : public PacketcaptureBase {
  std::string random_number;
  std::chrono::system_clock::time_point timeP;
  std::chrono::nanoseconds temp_offset;
+ std::string filter;
+ uint32_t snaplen = 262144;  /* 65535 for no sliced packets */
```
DavideAG: please use the snaplen in globalheader. there's no reason to duplicate the code

src/services/pcn-packetcapture/src/Packetcapture.cpp
```diff
@@ -435,4 +376,65 @@ void Packetcapture::to_timeval(std::chrono::system_clock::duration& d, struct ti
...
+}
+
+void Packetcapture::filterCompile(std::string str, struct sock_fprog * cbpf) {
+    int ret,i;
+    struct bpf_program _bpf;
+    const struct bpf_insn *ins;
+    struct sock_filter *out;
+    int link_type = 1;   //ETHERNET type
+
+    ret = pcap_compile_nopcap(65535, link_type, &_bpf, str.c_str(), 1, 0xffffffff);
```
DavideAG: It looks like you're forcing a 65535 snaplen here.
That way you won't be able to change it.

frisso: In this case, I feel Riccardo is correct. The *classical* eBPF code returns different values based on the snaplen, which is used to tell Linux how many bytes we have to return to user, so in cBPF you have to generate a new filter each time you update the snaplen.
In *eBPF*, the amount of data you return to user does not depend on the BPF code, but has to be enforced in some other way, e.g., by copying only the initial part of the packet to the control plane. So, we do not need to re-generate the (cBPF) code each time the snaplen is changed, and we can stay with a "default" value instead.

frisso:
Riccardo, please rebase your code with master. It looks there are a lot of files that look modified, but that you didn't touch at all. This makes the review (and not only the review) really hard.

src/services/pcn-packetcapture/src/Packetcapture.cpp
DavideAG: Thank you for your comment @frisso . So we can let this value as fixed.

then perhaps you're referring to this
https://github.com/polycube-network/polycube/pull/273#discussion_r382865408

frisso: Yes, correct. Also #273  does not need to be done imho.

src/services/pcn-packetcapture/src/Packetcapture.cpp
ricjames11: you mean the  if (CapStatus == 1) ?

Documentation/services/pcn-packetcapture/packetcapture-filter.rst
```diff
@@ -0,0 +1,65 @@
...
+generate the C code
+
+For reference see: `cloudflare project <https://blog.cloudflare.com/xdpcap/>`__
+
+
+Example of C code generated
+---------------------------
+Here is the generated C code for the filter "icmp":
+
+::
```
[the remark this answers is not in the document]
ricjames11: Do I have to put also some comment in the code or just the code well formatted?

src/services/pcn-packetcapture/src/Packetcapture.cpp
ricjames11: thanks, I've changed, now I use the globalheader class

Documentation/services/pcn-packetcapture/packetcapture-filter.rst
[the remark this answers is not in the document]
frisso: Just formatting.

src/libs/polycube/include/polycube/services/bcc_exception.h
```diff
@@ -0,0 +1,60 @@
+/*
```
frisso: Really not sure we need to add the following files from bcc.

src/services/pcn-packetcapture/datamodel/packetcapture.yang
```diff
@@ -43,40 +43,18 @@ module packetcapture {
   leaf networkmode {
     type boolean;
     default false;
-    description "operative mode";
+    description "operating mode";
```
frisso: Capital letters: "Operating mode"

src/services/pcn-packetcapture/src/cbpf2c.cpp
```diff
@@ -0,0 +1,507 @@
...
+                    }
+                }
+            }
+        }
+    }
+    ret[sizeof(ret) - 1] = '\0';
+    return ret;
+}
+
+const char *cbpf2c::cbpf_dump_linux_k(uint32_t k) {
```
frisso: Can you please add a comment about this function?

ricjames11: done

PR #275, opened 2020-02-18
New transparent service: Dynamic Monitor

michelsciortino:
This PR adds a new Polycube transparent service: pcn-dynmon.
This service represents a VNF able to receive his dataplane's eBPF code through his REST interface and inject it at runtime.

The ebpf code can be accompanied by a description of the eBPF maps that it uses, in order to export their content through the REST interface as metrics.

The exported metrics can be retrieved in two formats:
- the `/metrics` endpoint returns the metrics as json and can represent all the complex structures used in the creation of the eBPF maps.

- the `/open-metrics` endpoint returns the metrics as plain-text with the OpenMetrics format and can represents only simple data types; this endpoint can be used as target in Prometheus in order to monitor the metrics exported by this service.


frisso:
@michelsciortino This look very good to me. Just make some "squash commit" in order to avoid to show your latest commits, which simply modify what was done before.
Finally, can you please make a "rebase" and commit again, so that we can merge the code once we make sure it is already aligned with the latest master? Thanks!

PR #285, opened 2020-04-01
Adding some hints in Polycube Documentation

michelsciortino:
This PR adds some hints on how to compile a new Polycube service and how services can be loaded and unloaded at runtime without shutting down the Polycube  daemon.


Documentation/developers/hints.rst
```diff
@@ -1,5 +1,90 @@
...
+
+    cmake ..
+
+3. compile the service to build a shared library object:
+
+  ::
+
+    make -j $(getconf _NPROCESSORS_ONLN)
+
+At this point the service has been compiled and the built shared library object can be found into the ``build/src`` folder named as ``libpcn-service_name.so``
```
palexster, line 57: Hi @michelsciortino , thanks for this PR. I would add here a sentence like to better connect the compilation part with the following part of the document: "Your cube, completely encapsulated in the .so file, is now ready to be loaded by the Polycube"

Documentation/developers/hints.rst
```diff
@@ -1,5 +1,90 @@
...
+
+Loading/unloading Polycube services at runtime
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Polycube services can be loaded and unloaded at runtime by using ``polycubectl``, without the necessity of shutting down the `polycubed` daemon. This greatly facilitates the debugging of your service, as you can simply load the updated version of your service without closing and restarting `polycubed`.
+
+Loading a service
+*****************
+In order to load a new service into Polycube a shared library object must be provided.
```
palexster, line 65: Suggestion. I would rephrase in:  In order to load a new service, we should provide to Polycube daemon the shared library object (.so) we compiled before.

Documentation/developers/hints.rst
```diff
@@ -1,5 +1,90 @@
...
+
+  ::
+
+    polycubectl services add type=lib uri=/absolute/path/to/libpcn-service_name.so name=service_name
+
+- the ``uri`` parameter indicates the absolute path to the shared library object of the service;
+- the ``name`` parameter indicates the name that will be use in the Polycube rest API for the service.
+
+After having loaded the service, it can be istantiated as a normal service using ``polycubectl``.
+
```
palexster, line 78: I would add few commands to verify that the service has been correctly verified (e.g.; Create a new instance)

frisso:
Michel: first, thanks for this patch.
Two questions.
First, can you update "istance" with "instance"? :-)
Second, does this procedure works, or we have to put our hands to some of the Makefiles of the services?

michelsciortino:
Hi, sorry for the late response.
The stand-alone compilation works for any new service but does not work with every already existing service; this is due to the fact some of them make a explicit inclusion of the some libraries headers (`spdlog`) contained in the main Polycube source code folder, so they cannot be compiled out of that context.

Existing services stand-alone compilable:
- pcn-dynmon
- pcn-firewall
- pcn-packetcapture
- pcn-router
- pcn-synflood

Existing services NOT stand-alone compilable:
- pcn-bridge
- pcn-ddosmitigator
- pcn-helloworld
- pcn-iptables
- pcn-k8sfilter
- pcn-k8switch
- pcn-loadbalancer-dsr
- pcn-loadbalancer-rp
- pcn-nat
- pcn-pbforwarder
- pcn-simplebridge
- pcn-simpleforwarder
- pcn-transparent-helloworld

In order to make every existing service stand-alone compilable modifying the Makefiles is not enough as the linked libraries should be physically installed in the host along with their headers.

Anyway this can be managed by editing all the aforementioned services deleting the missing dependencies or by adding the needed source files in the services source folders.

frisso:
I feel that `spdlog` should not be needed in Polycube services; we can investigated more about this. Anyway, let's move on, this is ok for the time being.
Thanks, Michel!

PR #288, opened 2020-04-22
[WIP] Dynmon v1.1 - Supporting the egress dataplane path

michelsciortino:
Since in version 1.0 of Dynmon only the ingress dataplane path was supported, this PR adds the support for the execution of distinct eBPF codes on the ingress and egress dataplane paths.

This allows the generation and the exportation of distinct metrics upon the incoming and outgoing traffic of the service.


frisso:
Michel, please sync with @FedeParola as he just added the support for XDP egress, which is currently emulated (it doesn't exist right now in Linux). So, just make sure we're not duplicating the effort or creating some redundancy.

PR #292, opened 2020-04-26
Send packets to controller from the running program

### LINKED ISSUE #281: interface won't be reached with packetcapture cube attached
## Describe the bug
After attach a packetcapture cube to an host interface, the interface won't be reached from other hosts

## To Reproduce
1. create packetcapture cube with the following command
    polycubectl packetcapture add pc1 capture=bidirectional
    polycubectl pc1 set filter="host 4.4.4.2" (4.4.4.2 is ip address of another host connecting to the the host polycubed is running)
    polycubectl pc1 set dump="/tmp/jpi"
2. attach the cube to one of host interface, in my case interface name is "ens9" with ip "4.4.4.12"
   polycubectl attach pc1 ens9
3. ping ens9 from host 4.4.4.2
4. ens9 is not able to reach

## Expected behavior
Before attaching the cube to the interface, ens9 is reachable from other hosts; with cube enable it should be keep the same behavior 

## Please tell us about your environment: 
1. Ubuntu 18.04.1 LTS
2. 5.2.0-050200-generic
3. polycube version information
polycubectl:
 version: -128-NOTFOUND+ [git: (branch/commit): master/f199ac1b]
polycubed:
 version: -128-NOTFOUND+ [git: (branch/commit): master/f199ac1b]
4. Polycube Log: Attach the log from `/var/log/polycube/polycubed.log`

## Additional context
dumped packet information
> tcpdump -r /tmp/jpi.pcap -n
12:09:22.570537 IP 4.4.4.12 > 4.4.4.2: ICMP echo reply, id 18252, seq 4, length 64
12:09:22.570582 IP 4.4.4.12 > 4.4.4.2: ICMP echo reply, id 18252, seq 1, length 64
12:09:22.570626 IP 4.4.4.12 > 4.4.4.2: ICMP echo reply, id 18252, seq 3, length 64
12:09:22.570682 IP 4.4.4.12 > 4.4.4.2: ICMP echo reply, id 18252, seq 2, length 64
12:09:22.570727 IP 4.4.4.12 > 4.4.4.2: ICMP echo reply, id 18252, seq 4, length 64
12:09:22.570771 IP 4.4.4.12 > 4.4.4.2: ICMP echo reply, id 18252, seq 1, length 64
12:09:22.570815 IP 4.4.4.12 > 4.4.4.2: ICMP echo reply, id 18252, seq 3, length 64

looks the host replies the icmp request indefinitely

FedeParola:
This PR allows to send packets to the controller from the running program and then continue with the processing.
`pcn_pkt_controller()` now returns the return value of `bpf_perf_event_output()` helper.
Formerly packets were sent to the controller through a dedicated program executed via tail call and they could not be further processed in the datapath.
Fixes #281.

In details:

- The code of the two programs (`CTRL_TC_TX` and `CTRL_XDP_TX`) responsible of sending packets to the controller has been moved from the `Controller` class (`controller.cpp`) to the `pcn_pkt_controller()` helper in classes `CubeTC` (`cube_tc.cpp`), `CubeXDP` (`cubec_xdp.cpp`), `TransparentCubeTC` (`transparent_cube_tc.cpp`) and `TransparentCubeXDP` (`transparent_cube_xdp.cpp`) with minor modifications.
Standard and transparent versions of this code differ in the fact that standard code uses the `port_id` filed of metadata to store the port, while transparent version uses it to store the direction of the packet.
- `Controller` class now only instantiate two perf rings used to send packets to user plane (`controller_tc` and` controller_xdp`), declared as public tables, and then polls on them to retrieve the packets and forward them to the corresponding cube.
The code responsible of sending packets back to the fast path was not modified.
- `pcn_pkt_controller_with_metadata_stack()` helper has been removed since the calling program can send the packet to the controller and then let it proceed to the stack returning `RX_OK`.
- All services using the `pcn_pkt_controller()` helper (except from `packetcapture`) have been updated to call the helper and then return `RX_DROP`, in order to keep the same behavior as before.
The `packetcapture` returns RX_OK to let the packet pass.


frisso:
This should make PR #270 obsolete.

frisso:
@FedeParola, given the rather large amount of code that is involved in this PR, I would suggest to briefly explain how you changed the code, even here as a comment in this PR. Thanks!

FedeParola:
> @FedeParola, given the rather large amount of code that is involved in this PR, I would suggest to briefly explain how you changed the code, even here as a comment in this PR. Thanks!

Added more details in the PR description, let me now if there is still something not clear.

goldenrye:
@FedeParola a little late to review the the code, does this change fixes the problem or only refactoring the code? 
"Formerly packets were sent to the controller through a dedicated program executed via tail call and they could not be further processed in the datapath."
I am arguing with this - in existing design, the packet can still be processed if the dedicated program return value 0.

FedeParola:
> @FedeParola a little late to review the the code, does this change fixes the problem or only refactoring the code?

If you mean the problem in issue #281, yes, it is fixed.
There is still the problem that I mentioned in my comment to that issue though.
When a packet goes to the controller and then is sent back to the fast path through `send_packet_out()`, the ingress interface seen by the kernel is different from the original one (it is either `pcn_tc_cp` or `pcn_xdp_cp`) and this can cause problems.

> "Formerly packets were sent to the controller through a dedicated program executed via tail call and they could not be further processed in the datapath."
> I am arguing with this - in existing design, the packet can still be processed if the dedicated program return value 0.

You are right, the packet can still be processed by the network stack. With that sentence I meant that the packet can not be further processed by the ebpf program, unless you send it to the controller and then send it back to the fast path, with a decrease in performance and the problem described above.

PR #298, opened 2020-05-08
Reduce Polycube overhead

### LINKED ISSUE #300: [BUG] Deadlock in bridge service
## Describe the bug
In bridge service a deadlock may occur when deleting a port while a packet is being processed in the slowpath.
Mutex `ports_mutex_` is acquired when removing a port https://github.com/polycube-network/polycube/blob/2e4f4e2081eca0322e4729434aff897a6cac84fa/src/services/pcn-bridge/src/Bridge.cpp#L154-L158
This function then calls function `Cube<PortType>::remove_port()` that acquires `cube_mutex` https://github.com/polycube-network/polycube/blob/2e4f4e2081eca0322e4729434aff897a6cac84fa/src/libs/polycube/include/polycube/services/cube.h#L153-L157

Packets coming from the fast path are processed acquiring those mutex in reverse order.
The callback acquires cube_mutex https://github.com/polycube-network/polycube/blob/2e4f4e2081eca0322e4729434aff897a6cac84fa/src/libs/polycube/include/polycube/services/cube.h#L69-L86
and method `broadcast()` of the bridge then acquires `ports_mutex_` https://github.com/polycube-network/polycube/blob/2e4f4e2081eca0322e4729434aff897a6cac84fa/src/services/pcn-bridge/src/Bridge.cpp#L111-L120
Causing a deadlock.

## To Reproduce
This is difficult to reproduce since it depends on timing.
I found it while running test `pcn-bridge/tests/vlan/test1.sh` on my PR #298

## Expected behavior
No deadlock should occur.

## Please tell us about your environment: 
1. OS details: Ubuntu 18.04 
2. Kernel details: 5.3.0

## Additional context
I think `ports_mutex_` can be safely removed.
As far as I understand both the mutex have the same function: avoid deleting a port while processing a packet from the fastpath

FedeParola:
This series of patches aims to reduce the overhead introduced by Polycube.
Following optimizations will be applied:

- [x] In the program receiving a packet on an interface, directly inject the index and possible port of the first program to call, instead of retrieving it from a map. Every time a new porgram/port is attached/connected to the interface the rx program of that interface is re-injected with update parameters.
Commit 9127ad4
- [x] In XDP cubes, directly pass the pointer to `struct pkt_metadata` contained in the `PERCPU_ARRAY_MAP`, instead of copying it to a local structure and then copying it back to the map.
Commit 44d58d2
- [x] Dynamically generate the `pcn_pkt_redirect()` function of standard cubes with a switch-case construct containing the action to perform for each port, instead of relying on a map.
When a packet is directed out of an interface directly call the `bpf_redirect()` helper instead of calling an additional program that performs this task (`ExtIfaceTC::TX_CODE` and `ExtIfaceXDP::TX_CODE`).

Fixes  #300

Here are some numbers that show basic forwarding performance of Polycube (the used program simply receives the packet on an interface, swaps src and dst macs and forwards it on the other interface).
Single core tests run on a Intel Xeon @2.6GHz.
```
Pure XDP   7.07 Mpps
Polycube            CURRENT       OPTIMIZED
XDP_DRV           4.45 Mpps       6.41 Mpps
TC                1.33 Mpps       1.66 Mpps

### DIFF
--- a/src/libs/polycube/include/polycube/services/cube_iface.h
+++ b/src/libs/polycube/include/polycube/services/cube_iface.h
@@ -84,7 +84,8 @@ class CubeIface : virtual public BaseCubeIface {
   virtual void remove_port(const std::string &name) = 0;
   virtual std::shared_ptr<PortIface> get_port(const std::string &name) = 0;
 
-  virtual void update_forwarding_table(int index, int value) = 0;
+  virtual void update_forwarding_table(uint16_t port, uint32_t next,
+                                       bool is_netdev) = 0;
 
   virtual void set_conf(const nlohmann::json &conf) = 0;
   virtual nlohmann::json to_json() const = 0;
@@ -98,7 +99,7 @@ class CubeIface : virtual public BaseCubeIface {
 
 class TransparentCubeIface : virtual public BaseCubeIface {
  public:
-  virtual void set_next(uint32_t next, ProgramType type) = 0;
+  virtual void set_next(uint16_t next, ProgramType type, bool is_netdev) = 0;
   virtual void set_parameter(const std::string &parameter,
                              const std::string &value) = 0;
   virtual void send_packet_out(const std::vector<uint8_t> &packet, Direction direction,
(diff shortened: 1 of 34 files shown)


FedeParola:
> so we still cannot support XDP egress?

The XDP egress hook support has not been merged in the kernel yet, but I think it will soon be: https://lore.kernel.org/netdev/20200427224633.15627-1-dsahern@kernel.org/#t

PR #302, opened 2020-05-13
Fix Missing `bc` command

s41m0n:
This commit fixes a previous PR I made where, to check the OS versions, I used `bc` from command line assuming that it existed. It turned out that this caused problem, so now I am proposing to add this command in the installation process


s41m0n:
This PR should be reviewed as soon as possible, otherwise all build would fail due to `bc command not found`

PR #305, opened 2020-05-21
Updated BCC submodule to version 0.14.0

s41m0n:
This PR aims to update the submodule BCC in Polycube to the latest version
0.14.0, according to our recent updated in https://github.com/polycube-network/bcc/tree/polycube

### DIFF
--- a/src/services/CMakeLists.txt
+++ b/src/services/CMakeLists.txt
@@ -41,4 +41,4 @@ add_service(dynmon pcn-dynmon)
 # save string to create code that load the services
 SET_PROPERTY(GLOBAL PROPERTY LOAD_SERVICES_ ${LOAD_SERVICES})
 
-install(DIRECTORY datamodel-common DESTINATION include/polycube)
+install(DIRECTORY datamodel-common DESTINATION /usr/local/include/polycube)


s41m0n:
Do not merge please, the latest commit was to test the CI, but it has nothing to do with the PR

s41m0n:
Honestly, I do not understand why this last commit was needed, but I noticed that was the only option (as far as I've understood) to make datamodels be installed in /usr/local/include and not in /usr/include.. 
Installing from scratch works, if it is ok I would give this PR a shot to be merged

goldenrye:
> Honestly, I do not understand why this last commit was needed only now, but I noticed that was the only option (as far as I've understood) to make datamodels be installed in /usr/local/include and not in /usr/include..
> Installing from scratch works, if it is ok I would give this PR a shot to be merged

I run into the same issue and conclude that latest BCC introduces "set(CMAKE_INSTALL_PREFIX "/usr" CACHE PATH "path to install" FORCE)" which overwrites the default /usr/local install prefix directory. Removing that also works for me.

s41m0n:
> > Honestly, I do not understand why this last commit was needed only now, but I noticed that was the only option (as far as I've understood) to make datamodels be installed in /usr/local/include and not in /usr/include..
> > Installing from scratch works, if it is ok I would give this PR a shot to be merged
> 
> I run into the same issue and conclude that latest BCC introduces "set(CMAKE_INSTALL_PREFIX "/usr" CACHE PATH "path to install" FORCE)" which overwrites the default /usr/local install prefix directory. Removing that also works for me.

Thanks, I did not notice it and I supposed they had not changed something like this! I'll update our BCC removing these 3 lines in CMakeList.txt

PR #306, opened 2020-05-22
Improving Polycube doc - Install/Update kernel process

s41m0n:
Since we landed to Ubuntu 20.04, I though that a bit more specifics could be added into the guide


s41m0n:
Do you see anything else to be updated in the installation process @frisso ?

PR #311, opened 2020-06-10
Dynmon v1.1

s41m0n:
**New Dynmon v1.1 release**

This new version aims to introduce the following features :
* Map atomicity thanks to program swap before read (if specified by the user)
* Map deletion after read (if specified by the user)
* QUEUE/STACK map support in Polycube
* PERCPU_ARRAY map read support in Dynmon

## Map Atomicity

To ensure an atomic read of a BPF map, the user can specify the `swap-on-read` flag. When enabled, Dynmon creates a parallel code to be injected alternatively with the new one and read the map accordingly. However, the map used in the swapped code is a totally different map created by appending a specific suffix to all the occurrences. 

The new map declaration is copied also in the original program to ensure that when swapping back the code this map would not be destroyed. The same for the original map declaration in the swapped code.

The user is totally unaware of what happens in the lower layers. In fact, he still sees its original injected code when querying Dynmon, while the service loads/unloads the correct one (the original code is copied and modified as said before). 

## QUEUE/STACK
These maps introduce a new level of abstraction which is different that the previous ones, since the only admitted operation on Queues/Stacks are PUSH/POP (and PEEK but that's not the point). The fact is that the actual RawTable allows only to perform lookup, but this is not working with new data types.

My solution is the following:
* Added raw push/pop methods to RawTable
* Added utility functions to check the current kernel version, in order to disable at runtime these operations if not supported (even though the bpf compile would throw an error if you try to inject code with Queue/Stack in a running Polycube instance with kernel < 5.0.0)
* Added a QueueStackTable class, which is a template class extension of a RawTable (like all the other tables in Polycube)

I would have preferred making a RawTable class smarter, meaning that if a GET is called it check whether it has to perform a simple lookup or a pop, depending on the underlying map type. However, I don't know if you would like a solution like that, since it will introduce more "intelligence" to a Raw class.

Both pop/push operations has been introduced in the RawTable, but of course they would not work if the underlying table is not a queue/stack. This is up to the consumer to call the right method, if it is using such table. This addition allows Dynmon to interact with a RawTable, no matter its type, since it has to check a runtime both table and key type, and creating a specific table depending on the key type (int/struct ...) would not only make the code less comprehensible, but it would also be way bigger in size.

Finally, Dynmon has been refactored and enhanced, allowing also all PerCPU map types to be extracted in json format.

## PERCPU_ARRAY
This map type is not readable through RawTable interface. Instead of creating an entry of sizeof(leaf), Dynmon creates an entry of sizeof(leaf)*n_cpu. This way all the percpu values can be retrieved


s41m0n:
BEFORE MERGING, I need some answer (from Slack channel).

src/libs/polycube/src/common.cpp
```diff
@@ -108,4 +108,38 @@ PortStatus string_to_port_status(const std::string &status) {
...
+    }
+
+    if (vnum1 > vnum2)
+      return true;
+    if (vnum2 > vnum1)
+      return false;
+  }
+  return true;
+}
+
```
sebymiano, line 144: @s41m0n If I am not wrong, in Polycube utils there was already a function to check the current kernel version [1].
It would be better to use that one and avoid duplicate code.

https://github.com/polycube-network/polycube/blob/83b4a6d6720b6bf91caf63f3a89f9800f3b847e9/src/polycubed/src/utils/utils.cpp#L28

src/libs/polycube/src/table.cpp
```diff
@@ -36,13 +37,23 @@ class RawTable::impl {
...
 };
 
-RawTable::impl::~impl() {}
+RawTable::impl::~impl() {
+}
 
-RawTable::impl::impl(void *op) : fd_(*(int *)op) {}
+RawTable::impl::impl(void *op) : fd_(*(int *)op) {
+  auto version = get_kernel_release();
+  is_queue_stack_enabled = compare_kernel_release(version, QUEUE_STACK_KERNEL_RELEASE);
```
sebymiano, line 54: Honestly, I would prefer the other way, i.e., create a different map for the Queue and Stack.
I have a question, what is the behavior of the `bpf_map_lookup` with those tables? And, on the other hand, what is the behavior of the `bpf_map_lookup_and_delete` with the other tables?

src/libs/polycube/src/table.cpp
```diff
@@ -36,13 +37,23 @@ class RawTable::impl {
...
+  is_batch_enabled = compare_kernel_release(version, BATCH_KERNEL_RELEASE);
```
sebymiano, line 55: Did you add the batch support? If not, maybe it is better to add this change in that PR.

src/services/pcn-dynmon/src/Dynmon.h
```diff
@@ -6,13 +6,25 @@
 #include "models/DataplaneConfig.h"
 #include "models/Metric.h"
 #include "models/Metrics.h"
+#include "swap/SwapCompiler.h"
```
sebymiano, line 9: I would not call it a "compiler". A compiler is something else and this may confuse people. 
Let's call it `SrcCodeRewriter` or just `Rewriter`, it gives a better idea of what this component does.

src/libs/polycube/src/common.cpp
s41m0n: https://github.com/polycube-network/polycube/blob/83b4a6d6720b6bf91caf63f3a89f9800f3b847e9/src/polycubed/src/utils/utils.cpp#L28

Found, I'm going to use that one, thanks!

src/libs/polycube/src/table.cpp
s41m0n: Perfect, I will change the implementation according to your suggestion :)
According to https://github.com/torvalds/linux/blob/24085f70a6e1b0cb647ec92623284641d8270637/kernel/bpf/syscall.c#L1442 :
* `bpf_map_lookup_and_delete` returns ENOSUPP if the map is not Queue/Stack
* `bpf_map_lookup` seems not to work with Queue/stack (table.next() returns always -1)

src/libs/polycube/src/table.cpp
s41m0n: I haven't added it yet, but I tried to make it work in a personal branch, but I'm still figuring out how to correctly call the function (unfortunately there's no doc). Should I implement it in this PR?

src/services/pcn-dynmon/src/Dynmon.h
s41m0n: Agree, I'm going to change it asap :)

s41m0n: Solved in 523ff8b

src/libs/polycube/src/common.cpp
s41m0n: Solved in 523ff8b

src/libs/polycube/src/table.cpp
s41m0n: Solved in 523ff8b 
Let me know if this new solution is appreciated @sebymiano

src/libs/polycube/src/table.cpp
s41m0n: Solved in 523ff8b 
Unfortunately, batch works only with primitive data types and with standard maps (not queue/stack). So I've added these new function to the RawTable class that can be used, but in Dynmon I've commented the code where I performed batch research since no useful by now

s41m0n: Maybe in the following kernel releases they will make it work, and then we will already have the code for the new extraction type :)

s41m0n:
Useful links:
* https://lore.kernel.org/bpf/20190829064512.2751289-1-yhs@fb.com/  batch examples
* https://github.com/torvalds/linux/blob/24085f70a6e1b0cb647ec92623284641d8270637/kernel/bpf/syscall.c#L1341 generic batch function
* https://docs.huihoo.com/doxygen/linux/kernel/3.7/include_2linux_2errno_8h.html linux error codes > 500
* https://github.com/torvalds/linux/tree/3d77e6a8804abcc0504c904bd6e5cdf3a5cf8162/kernel/bpf kernel release 5.7.0 where there are all the bpf files + batch
* https://www-numi.fnal.gov/offline_software/srt_public_context/WebDocs/Errors/unix_system_errors.html linux error codes < 124
* https://www.fsl.cs.sunysb.edu/kernel-api/re245.html put_user function used in get_batch (THIS IS THE CAUSE OF USING ONLY PRIMITIVE DATA TYPE)
* https://lore.kernel.org/bpf/20191119014357.98465-3-brianvv@google.com/ batch PR with some doc (the only one available online)

PR #314, opened 2020-07-02
Script to create zip and tar of Polycube with submodules

pinoOgni:
This PR introduce a script to create an archive (zip and tar) with all the polycube sources, including submodules. It is possible to choose to make an archive from a branch with a version or from a tag and use the tag for the version.


frisso:
@pinoOgni This looks really great!
@sebymiano Two comments.
1) I moved line "git log -1" into an "if/then/else" block, so to avoid this to be executed always. This is useful when you want to compile the sources from the source pack, when you may not have git installed.
2) I wonder if the 'install' should support one more option, such as "install everything" or "install only binaries". What do you think?

PR #316, opened 2020-07-13
New dockerfile version

s41m0n:
This PR tries to propose a solution to the Docker container size problem.
With this proposal, the image is built extracting from an intermediate layer only
the executables (already compiled) and libraries needed.

The difference between the older image and the new one  both decompressed is:
```bash
s41m0n/polycube            latest             5d265335b2ff        5 minutes ago       265MB
polycubenetwork/polycube   latest              bc806e9934ca        3 days ago          2.35GB
```

The image compressed size (you can check here https://hub.docker.com/repository/docker/s41m0n/polycube/tags?page=1 ) will be only 81.15 MB (compared to 761.79 MB).

Still a WIP since I am still checking that all the components are correctly working.
So far it seems ok (surprisingly).

**PS1**:
there is actually nothing in the resulting image (no sudo, ip, ifconfig, etc). Before merging, we will have to discuss whether to provide 2 different types of images (1 complete with utilities and 1 only for exec).

**PS2**:
organizing the Dockerfile this way would completely remove the polycubebot/base_image, since all the layers are built temporary. 
Pros:
* organized and in the same Dockerfile
* no need of an extra repository

Cons:
* depending on the layers, if you modify a row it'll probably recompile the entire polycube 


frisso:
This looks really great, @s41m0n !
Two minor comments.
1) The testing pipeline fails so far, no idea why
2) Actually, three Dockers were built in Polycube, one for the standard Polycube, the other for pcn-iptables, the final for pcn-k8s. Not clear the dependencies between them. Probably it would be nice to document them.

aasifshaikh:
- @s41m0n Awesome! Much more slimmer version from my last test.
Nice to see that we're removing base image dependency. Did you build with --squash or without? by chance, if we may benefit more on image size reduction.

A thorough testing is required to validate it works as expected.

s41m0n:
> This looks really great, @s41m0n !
> Two minor comments.
> 
>     1. The testing pipeline fails so far, no idea why
> 
>     2. Actually, three Dockers were built in Polycube, one for the standard Polycube, the other for pcn-iptables, the final for pcn-k8s. Not clear the dependencies between them. Probably it would be nice to document them.

Thanks! I think I have to debug a lot especially with the current pipeline...
Concerning the ci/dockercloud test, when has it been added? And what does it test?

s41m0n:
>     * @s41m0n Awesome! Much more slimmer version from my last test.
>       Nice to see that we're removing base image dependency. Did you build with --squash or without? by chance, if we may benefit more on image size reduction.
> 
> 
> A thorough testing is required to validate it works as expected.

Hi Asiif and thanks! I did not build with `--squash`, so maybe the final size would be even less. Surely we need to create some tests to check whether all components work as expected, this was just a first approach to test the feasibility :)

s41m0n:
I slightly modified the Dockerfile to fix some errors during the `pcn-k8s` image build and now all tests seems to pass.
I may use some help here (@aasifshaikh and whoever is available) to discuss whether all images are correctly built and running, in the sense that I have always used only the DEFAULT image, so I do not know if the `pcn-k8s` one (or ip-tables) needs some extra software installed in the docker.

Inside a Docker image there's only:

* polycubed
* polycubectl

All the other sw is removed, meaning that if someone wants to interact with CURL from inside the docker he has to install it (seems legit to me, why should someone interact from inside the docker?)

s41m0n:
Update: compiling with `--squash` did not help to improve the image size, it just merged everything into a single layer. 
I think that the max we can achieve is 265 MB size

s41m0n:
Dear all,
I can happily announce that also the Docker pcn-k8s has successfully passed the tests.
Thanks to Raffaele, we have successfully installed the CNI using my pcn-k8s docker, compiled with the new Dockerfile. Not only did the pods correctly start, but also all the containers were up and running with no errors.
I would say that we are close to merge this PR and finally have a lighter image.
Maybe we wait a few more days, to see if anyone has suggestions, then we can proceed.

PR #318, opened 2020-07-28
Ebpf/packet timestamp

s41m0n:
This PR aims to introduce an helper to ease packet's timestamp computation. 
Moreover, the new value will refer to the Unix Epoch instead of kernel timer.

**Usage**
Old style -> `ctx->tstamp == 0 ? bpf_ktime_get_ns() : ctx->tstamp` (worst case you have an independent kernel timer value)
New style: `pcn_get_packet_tstamp(ctx)` and you always have a Unix Epoch related value

**How it works**
When Polycubed starts, it is injected a special value as a flag to all eBPF programs. This value comes from the following equation:

```bash
epoch_at_y = epoch_at_x + (uptime_at_y - uptime_at_x)
```
This way the newly introduced helper will use this value to compute the new timestamp.
All the values retrieved at time X are computed in ControlPlane once. The uptime_at_y is retrieved by using `bpf_ktime_get_ns()`, which unfortunately gets the uptime without the sleep/hybernate/suspend time.
Another workaround would be inserting these values into a BPF_MAP and updating them every K seconds, to avoid that after a sleep the computed value is wrong. Since in the next kernel release there would be the possibility to call `bpf_ktime_get_ns()` considering all the time since boot, I though that this would not be the best solution (also every time you have to compute a timestamp you would have to lookup from that map).

**Known bug**
If the device running Polycube suspends/hybernates/sleeps when Polycubed is already running, all the following computed values will not consider that time, thus the computed timestamp will not be exact. Hopefully, as suggested by Federico Parola, in the next kernel release will be released a bpf helper to get kernel boot time considering also those periods of inactivity.

### DIFF
--- a/src/polycubed/src/base_cube.cpp
+++ b/src/polycubed/src/base_cube.cpp
@@ -21,12 +21,32 @@
 namespace polycube {
 namespace polycubed {
 
+/** Function to compute a base Unix Epoch timestamp to be used as reference by
+ * all the timestamp computed using pcn_get_packet_epoch(ctx).
+ *
+ * Supposing the packet arrives at time X and we precomputed Unix Epoch and
+ * kernel timer at time Y:
+ *
+ * Epoch_at_X = Epoch_at_Y + (Kernel_at_X - Kernel_at_Y)
+ *
+ * where Kernel_at_X is given by the function bpf_ktime_get_ns()
+ * */
+uint64_t genBaseTime() {
+  // getting system clock (to retrieve Unix Epoch)
+  auto epoch = std::chrono::system_clock::now();
+  // getting the same time as bpf_ktime_get_ns()
+  auto up = std::chrono::steady_clock::now();
+  // return value to be summed to packet's bpf_ktime_get_ns()
+  return epoch.time_since_epoch().count() - up.time_since_epoch().count();
+}
+
 std::vector<std::string> BaseCube::cflags_ = {
     std::string("-D_POLYCUBE_MAX_NODES=") +
         std::to_string(PatchPanel::_POLYCUBE_MAX_NODES),
     std::string("-D_POLYCUBE_MAX_BPF_PROGRAMS=") +
         std::to_string(_POLYCUBE_MAX_BPF_PROGRAMS),
     std::string("-D_POLYCUBE_MAX_PORTS=") + std::to_string(_POLYCUBE_MAX_PORTS),
+    std::string("-D_EPOCH_BASE=") + std::to_string(genBaseTime()),
 };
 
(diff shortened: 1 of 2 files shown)


src/polycubed/src/base_cube.cpp
```diff
@@ -21,12 +21,24 @@
 namespace polycube {
 namespace polycubed {
 
+uint64_t genBaseTime() {
```
frisso, line 24: I suggest to comment this code a little bit more.

src/polycubed/src/cube_tc.cpp
```diff
@@ -324,6 +324,15 @@ int pcn_vlan_push_tag(struct CTXTYPE *ctx, u16 eth_proto, u32 vlan_id) {
   return bpf_skb_vlan_push(ctx, eth_proto, vlan_id);
 }
 
+/* helper to get packet timestamp related to Unix Epoch */
```
frisso, line 327: If I am not wrong, the code in XDP and TC looks the same.
Is this wanted? In this case, shouldn't we write it once, maybe in the base cube?

FedeParola: Have you tested the xdp version? There shouldn't be a tstamp field in `struct xdp_md`: https://github.com/torvalds/linux/blob/master/include/uapi/linux/bpf.h#L3754-L3763

src/polycubed/src/base_cube.cpp
FedeParola: I agree, why are you using two different clocks?

s41m0n: Comments + DOC added. I fixed it, the second timer was a mistake

src/polycubed/src/cube_tc.cpp
s41m0n: I decided to comment the `ctx->tstamp` selection since its value changes too many times. To compute the Unix epoch, I need the kernel timer, thus `bpf_ktime_get_ns()`. I cannot rely on `ctx->tstamp`. For those use cases where the timestamp does not have to be Unix epoch, it can still be used thus.

s41m0n:
I confirm that this latest implementation works also in XDP mode since I do not use anymore ctx->tstamp value, but the precomputed one + bpf_ktime_get_ns()

FedeParola:
Just one last thing, wouldn't it be better to rename the helper in something like `pcn_get_system_time()` or `pcn_get_time_epoch()` since we are not retrieving the timestamp of the packet? (of course the value we get can be used to timestamp the packet, but also for other tasks)

Issue #277, opened 2020-03-02
[Feature Request] Exploit OS timestamping when capturing packets


frisso:
## Problem statement
When capturing packets, e.g., in the packetcapture service, the dataplane code has to take the timestamp of the packet. However, it seems that somewhere the Linux kernel handles the timestamp of the packet, given that there is a `tstamp` member in the `sk_buff` structure.

## Describe the solution you'd like
It would be nice to be able to take the value possibly stored in `sk_buff.tstamp` instead of taking the timestamp in eBPF dataplane code.

## Additional context
From some tests done on the packetcapture service, it seems that the value in `sk_buff.tstamp` is always zero (program running in TC, attached to both physical and virtual interfaces). However, the member is there and if we type `man packet`, which refers to `AF_SOCKET`, it says that the `PACKET_TIMESTAMP` is just for this purpose. So, it is not clear how to exploit this timestamping in eBPF code.

s41m0n:
The proposal in #320 is to provide a method to generate a Unix epoch timestamp by simply calling `bpf_ktime_get_ns()` and adding this value to a precomputed Epoch timestamp.

The problem we are facing with `sk_buff.tstamp` is that sometimes the field is empty or contains an unknown value.
Probably I am missing something, but right now it seems that we cannot fully trust this field, even because it does not exists in a  XDP program

PR #320, opened 2020-07-30
fix ebpf timestamping nomenclature

s41m0n:
PR to fix name of Polycube helper to retrieve timestamp in Unix Epoch + Doc

Related issue: #277


frisso:
Great!
Many thanks, Simone!

PR #325, opened 2020-08-31
docker github action test

acloudiator:
Docker github action to test the build based with event trigger

### DIFF
--- a/.github/workflows/docker-image.yml
+++ b/.github/workflows/docker-image.yml
@@ -1,0 +1,18 @@
+name: Docker Image CI
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Build the Docker image as Default mode 
+      run: docker build . --file Dockerfile --tag default-github-action-test:$(date +%s)
--- a/.github/workflows/docker-image.yml
+++ b/.github/workflows/docker-image.yml
@@ -7,7 +7,6 @@ on:
     branches: [ master ]
 
 jobs:
-
   build:
 
     runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -9,7 +9,8 @@ FROM ubuntu:18.04 AS builder
 
 ARG DEFAULT_MODE=default
 ENV MODE=$DEFAULT_MODE
-RUN --mount=target=/polycube cp -r /polycube /tmp/polycube && cd /tmp/polycube && \
+COPY /polycube/ /tmp/polycube
```
s41m0n, line 12: Shouldn't it be `COPY . /tmp/polycube/`? I'm scared that `COPY /polycube/ ...` fails because, when building, the process is already inside the directory

.github/workflows/docker-image.yml
```diff
@@ -0,0 +1,21 @@
...
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Build the Docker image as Default mode 
+      run: |
+        pwd
+        cd ..
+        ls -ltr
```
s41m0n, line 20: Try also by adding `export DOCKER_BUILDKIT=1` before building

acloudiator: Buildkit doesn't seems to be supported by github action. We need to get rid of this mount, last time I checked it was an experimental feature.

s41m0n: I found this discussion https://github.com/docker/github-actions/issues/12#issuecomment-600540752 where somebody said that DOCKER_BUILDKIT is supported on Github Actions. I do not know if it is true or not, but maybe we could try to insert `env: DOCKER_BUILDKIT:1 ` as reported in this yml file https://github.com/docker/github-actions/blob/826b584752ffa6abc1a2d94da26f348e3e4f73d0/.github/workflows/main.yml#L26

frisso:
Superseded by #364

PR #332, opened 2020-09-16
Added a more complete example

frisso:


goldenrye (approved):

s41m0n (approved):

Documentation/services/pcn-firewall/firewall.rst
```diff
@@ -35,15 +35,23 @@ How to use
 Ingress ad egress chains
 ^^^^^^^^^^^^^^^^^^^^^^^^
 
-The service supports independent ingress and egress policy chains, with two different policy sets.
+The service supports independent _ingress_ (packets that are try to reach your TCP/IP stack) and _egress_ (pac policy chains, with two different policy sets:
+  - **ingress**: packets that come from the external world and that are try to reach the _inside_ of your network;
+  - **egress**: packets that come from your inside network and that are trying to reach the external world.
```
FedeParola, line 40: If I'm not wrong ` must be used on ReadTheDocs to create italic text instead of \_
```suggestion
The service supports independent `ingress` and `egress` policy chains, with two different policy sets:
  - **ingress**: packets that come from the external world and that are trying to reach the `inside` of your network;
  - **egress**: packets that come from your inside network and that are trying to reach the external world.
```

Documentation/services/pcn-firewall/firewall.rst
```diff
@@ -35,15 +35,23 @@ How to use
...
 
 Rule insertion
 ^^^^^^^^^^^^^^
 
 Rule insertion is guaranteed to be *atomic*: during the computation of the new datapath, the old rule set is used until the new rule set is ready, and only at that moment the new policies will be applied.
 
-Rule insertion is an expensive operation. For this reason, we have thought about different endpoints to optimize expensive operations:
+Rules can be:
+  - **inserted**: the ``insert`` action adds your rule at the _beginning_ of the ruleset (i.e., it becomes the new rule ``0``; existing rules are pushed down).
+  - **appended**: the ``append`` action adds your rule at the _end_ of the ruleset (i.e., it becomes the last rule of the ruleset, before the _default_ rule).
```
FedeParola, line 50:
```suggestion
  - **inserted**: the ``insert`` action adds your rule at the `beginning` of the ruleset (i.e., it becomes the new rule ``0``; existing rules are pushed down).
  - **appended**: the ``append`` action adds your rule at the `end` of the ruleset (i.e., it becomes the last rule of the ruleset, before the `default` rule).
```

FedeParola (changes requested):

Documentation/services/pcn-firewall/firewall.rst
frisso: Thanks for checking my the typos.
With respect to italic, ReadTheDocs supports "_" as a marked for italic, so I'm going to reject your suggestions in that respect. Thanks!

palexster (approved):

Issue #334, opened 2020-09-20
[Feature Request] Firewall "FORWARD" policy should be renamed as "ACCEPT"


frisso:
Given that the firewall is a transparent service, it doesn't forward anything; in fact, the forwarding is possibly done by the service the firewall is attached to.
So, for clarity it should be better to rename the FORWARD action into ACCEPT.

FedeParola:
Added with https://github.com/polycube-network/polycube/pull/337

PR #337, opened 2020-09-21
modified FORWARD policy name to ACCEPT and enabled by default ESTABLISHED

### LINKED ISSUE #331: [BUG] accept-established in firewall should be ON by default
The documentation and the YANG model says that the `accept-established` flag should be ON. Instead, it is OFF by default.
This is annoying, as _established_ sessions are not accepted by default by the firewall.

To see the problem, use the following commands:

```
polycubectl add firewall fw1
polycubectl attach fw1 enp0s3
polycubectl show fw1
```

Now, you can see that the flag is OFF:
```
fulvio@fulvio-VirtualBox:~$ polycubectl fw1 show
name: fw1
uuid: 5abf419a-9268-4189-9da4-31aa26a82cf0
service-name: firewall
type: TC
loglevel: INFO
parent: enp0s3
conntrack: on
accept-established: off
                    ^^^

```

s41m0n:
FIX #331
ENHANCE #334

Tried quickly with docker `s41m0n/polycube:fulvio`, if someone can help testing  this and modifying all the tests accordingly would be amazing.


src/services/pcn-firewall/datamodel/firewall.yang
```diff
@@ -92,8 +92,8 @@ module firewall {
     leaf action {
       type action;
       polycube-base:init-only-config;
-      description "Action if the rule matches. Default is DROP.";
-      polycube-base:cli-example "DROP, FORWARD, LOG";
+      description "Action if the rule matches. Default is ACCEPT.";
```
goldenrye, line 95: Default action should be DROP not ACCEPT

src/services/pcn-firewall/datamodel/firewall.yang
```diff
@@ -182,8 +182,8 @@ module firewall {
 
     leaf default {
       type action;
-      description "Default action if no rule matches in the ingress chain. Default is DROP.";
-      polycube-base:cli-example "DROP, FORWARD, LOG";
+      description "Default action if no rule matches in the ingress chain. Default is ACCEPT.";
```
goldenrye, line 185: same as above

src/services/pcn-firewall/datamodel/firewall.yang
frisso: This was changed recently because of a logistical problem.
Let's assume you deploy a firewall on a remote machine, and the default action is DROP. What happens, is that as soon as you start the firewall, you loose the connection to the remote machine.
So, to avoid this problem (very important in real deployments), we changed the default action to ACCEPT. Obviously, the user can change it to DROP upon its preferences, hopefully after having properly configured the ruleset.

goldenrye:
Actually we can first configure the rule allowing the ssh traffic before
attaching the firewall to the interface.

src/services/pcn-firewall/datamodel/firewall.yang
goldenrye: @frisso A better practice without changing the default policy is 
1) create firewall cube 
2) configure the firewall with rule accepting the management traffic 
3) attach the firewall cube to the interface

FedeParola: If we want to use iptables as an example of firewall behaviour I think that ACCEPT is the default action (for example after a fresh Ubuntu server installation)

frisso: @goldenrye What you say is definitely safer, but in our experience this would trigger a lot of problems, because many people would  first start and attach the firewall, then configure it. So, they would be cut out of their environment because the default 'deny' rule. I would prefer to stay with a default "accept" and improve the documentation in order to warn about this problem. Btw, we can explicitly state that the preferred configuration choice, if you want to have a default 'deny' rule, is to configure the rule, then attach the firewall to the ports.

goldenrye (approved):

FedeParola (approved):

s41m0n:
It fails to execute Line 48 from Dockerfile, giving "No such lib*.so":
```
# copying polycube services
COPY --from=builder /usr/lib/lib*.so /usr/lib/
```
Has something changed in the configuration of the Jenkins' nodes?

s41m0n:
++++TEST ./../src/services/pcn-firewall/test/ping/test_ping_5.1.sh Failed++++
cause: unknown

++++TEST ./../src/services/pcn-firewall/test/tcp/test_tcp_4.sh Failed++++
cause: netcat connection time out

@FedeParola Can you help checking these two tests please? This is taking more time than expected

FedeParola:
I may have found the cause of the problem @s41m0n.
Jenkins slaves use a old version of the kernel (4.15) that doesn't have support for bounded loops and has the limit of 4096 instructions per ebpf program. To overcome the first problem the firewall performs loop unrolling and apparently the number of rules of the test are causing the size of the program to exceed the limit (at least this is what happens when I downgrade the kernel on my machine). Could you try lowering that number? In my tests 2000 rules should be safe

FedeParola:
Maybe I found out why the tests are passing on the master: original tests use a batch file to carry rules and then pass this file to the polycubectl command, however probably this method doesn't work and no instruction is added at all, leaving only the default (forward) policy.

s41m0n:
> I may have found the cause of the problem @s41m0n.
> Jenkins slaves use a old version of the kernel (4.15) that doesn't have support for bounded loops and has the limit of 4096 instructions per ebpf program. To overcome the first problem the firewall performs loop unrolling and apparently the number of rules of the test are causing the size of the program to exceed the limit (at least this is what happens when I downgrade the kernel on my machine). Could you try lowering that number? In my tests 2000 rules seem to be safe.
> 
> I still don't understand why the same tests passed formerly... But it is worth giving a try

The number of rules injected is the same, it hasn't changed. 
Even when there was the "transactional" mode instead of batch rule injection we used to insert more than 4000 rules

s41m0n:
> Maybe I found out why the tests are passing on the master: original tests use a batch file to carry rules and then pass this file to the polycubectl command, however probably this method doesn't work and no instruction is added at all, leaving only the default (forward) policy.

I remember to have tested file injection with "polycubectl" with Fulvio and it seemed to work. Anyway, now the rules are correctly inserted, I am able to see them all, and locally the test passes :(

FedeParola:
> > Maybe I found out why the tests are passing on the master: original tests use a batch file to carry rules and then pass this file to the polycubectl command, however probably this method doesn't work and no instruction is added at all, leaving only the default (forward) policy.
> 
> I remember to have tested file injection with "polycubectl" with Fulvio and it seemed to work. Anyway, now the rules are correctly inserted, I am able to see them all, and locally the test passes :(

Yes you are right, I didn't check too accurately, sorry.

> > I may have found the cause of the problem @s41m0n.
> > Jenkins slaves use a old version of the kernel (4.15) that doesn't have support for bounded loops and has the limit of 4096 instructions per ebpf program. To overcome the first problem the firewall performs loop unrolling and apparently the number of rules of the test are causing the size of the program to exceed the limit (at least this is what happens when I downgrade the kernel on my machine). Could you try lowering that number? In my tests 2000 rules seem to be safe.
> > I still don't understand why the same tests passed formerly... But it is worth giving a try
> 
> The number of rules injected is the same, it hasn't changed.
> Even when there was the "transactional" mode instead of batch rule injection we used to insert more than 4000 rules

Ok then the overflow is probably caused by the DEBUG instructions, try instantiating the firewall with no loglevel (default=INFO) and the problem should be gone.

s41m0n:
@FedeParola thanks, the debug option seemed to alter some test, it worked.

frisso (approved):

PR #341, opened 2020-09-27
ddosmitigator: handle insertion of already present address

FedeParola:
When an address is already present no operation is performed.
Formerly the address was erroneously removed from the ebpf table.

### DIFF
--- a/src/services/pcn-ddosmitigator/src/Ddosmitigator.cpp
+++ b/src/services/pcn-ddosmitigator/src/Ddosmitigator.cpp
@@ -176,30 +176,32 @@ Ddosmitigator::getBlacklistSrcList() {
 
 void Ddosmitigator::addBlacklistSrc(const std::string &ip,
                                     const BlacklistSrcJsonObject &conf) {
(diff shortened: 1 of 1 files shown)


goldenrye (approved):
Even without the check the old code looks harmless, it is weird that the key will be removed from the hash table, it is a bpf_update_elem()

FedeParola:
> Even without the check the old code looks harmless, it is weird that the key will be removed from the hash table, it is a bpf_update_elem()

The problem lies in this `emplace`:
https://github.com/polycube-network/polycube/blob/8bd4c5c2a65a6bb63e5bcae45e852ecd5ee3fa6b/src/services/pcn-ddosmitigator/src/Ddosmitigator.cpp#L201-L202
It causes the execution of the destructor of the old `BlacklistSrc` or `BlacklistDst` entry and the destructor removes the value form the ebpf map.

FedeParola:
I don't know what the `ci/dockercloud` test is about? How can I check why it is failing?

acloudiator (approved):
LGTM

PR #356, opened 2020-12-09
fix PERCPU maps entries deletion and improved injector python script

s41m0n:
when deleting entries in PERCPU_HASH map, the bucket was not zero-ed. Thus, following entries using that bucket would have junk values belonging to the previous entry.


FedeParola (approved):

frisso (approved):

frisso:
Great work, Simone: Thanks!

palexster (approved):

PR #357, opened 2020-12-10
Fix map reuse when reloading TC egress program of XDP cubes

FedeParola:
There is a bug when reloading a program in the egress path of a XDP cube: XDP egress is emulated with two programs, a TC one to handle packets coming from the stack and a XDP one to handle packets coming from other XDP cubes. This programs should share the same maps in order to behave as one. There is a bug that causes these maps to differ after a reload. This patch fixes the problem.


FedeParola:
Moved on master to perform tests

frisso (approved):

frisso (approved):

frisso:
Nice spot, Federico: thanks!

palexster (approved):

Issue #352, opened 2020-11-03
[BUG] segmentation fault when del nat rule 


eTHaNeH:
## Describe the bug
Polycubed segmentation fault occurred when trying to delete rule from NAT

## To Reproduce

> banyan@banyan-pc:$ sudo systemctl start polycubed
[sudo] password for banyan: 
banyan@banyan-pc:~$ 
banyan@banyan-pc:~$ polycubectl nat add nat1
banyan@banyan-pc:~$ 
banyan@banyan-pc:~$ polycubectl nat1 show
name: nat1
uuid: e750ec5d-3aec-466a-b96c-e270c4f09900
service-name: nat
type: TC
loglevel: INFO
parent: 
rule:
 snat:
 masquerade:
  enabled: false
 dnat:
 port-forwarding:
banyan@banyan-pc:~$ 
banyan@banyan-pc:~$ polycubectl nat1 rule del
banyan@banyan-pc:~$ 
banyan@banyan-pc:~$ polycubectl nat1 show
Get "http://localhost:9000/polycube/v1/nat/nat1/": EOF
banyan@banyan-pc:~$ 
banyan@banyan-pc:~$ polycubectl nat1 show
Get "http://localhost:9000/polycube/v1/nat1/": dial tcp 127.0.0.1:9000: connect: connection refused
banyan@banyan-pc:~$ 
banyan@banyan-pc:~$ 
banyan@banyan-pc:~$ sudo systemctl status polycubed
● polycubed.service - polycube service
   Loaded: loaded (/lib/systemd/system/polycubed.service; disabled; vendor preset: enabled)
   Active: failed (Result: signal) since Tue 2020-11-03 15:43:33 CST; 29s ago
  Process: 3818 ExecStart=/usr/local/bin/polycubed (code=killed, signal=SEGV)
 Main PID: 3818 (code=killed, signal=SEGV)
Nov 03 15:42:57 banyan-pc polycubed[3818]: [2020-11-03 15:42:57.828] [polycubed] [info] service synflood loaded using libpcn-synflood.so
Nov 03 15:42:57 banyan-pc polycubed[3818]: [2020-11-03 15:42:57.854] [polycubed] [info] service packetcapture loaded using libpcn-packetcapture.so
Nov 03 15:42:57 banyan-pc polycubed[3818]: [2020-11-03 15:42:57.882] [polycubed] [info] service dynmon loaded using libpcn-dynmon.so
[...]

PR #383, opened 2021-06-30
A possible solution to fix bug #352

pinoOgni:
This PR fixes bug #352 

If you reset the rule_ variable (putting it to nullptr) in the delRule method, a subsequent command like ` polycubectl nat1 show` ends up calling the toJsonObject method to construct the JSON to be displayed to the user. Here we have this statement conf.setRule(getRule ()->toJsonObject ()); which, however, calls the toJsonObject() on a nullptr variable which causes the error.

The solution was to leave the deletion of rule_ in the delRule method in such a way as to also delete all the other rules (snat, dnat, port-forwarding) and then do as in the constructor, creating an "empty" rule_.

This solution also works for the case in which no rules have been added to the NAT and a user tries to do ` polycubectl nat1 rule del`.

Here there is an example:
```
netgroup@giuseppe:~$ polycubectl nat nat1
netgroup@giuseppe:~$ polycubectl nat1 show
name: nat1
uuid: 3e4c0102-a77a-47d8-a536-8c78178cea3e
service-name: nat
type: TC
loglevel: INFO
parent: 
rule:
 snat:
 masquerade:
  enabled: false
 dnat:
 port-forwarding:

netgroup@giuseppe:~$ polycubectl nat1 rule del
netgroup@giuseppe:~$ polycubectl nat1 show
name: nat1
uuid: 3e4c0102-a77a-47d8-a536-8c78178cea3e
service-name: nat
type: TC
loglevel: INFO
parent: 
rule:
 snat:
 masquerade:
  enabled: false
 dnat:
 port-forwarding:
 
netgroup@giuseppe:~$ polycubectl nat1 rule snat append internal-net=10.0.0.0/24 external-ip=1.2.3.5
id: 0

netgroup@giuseppe:~$ polycubectl nat1 show
name: nat1
uuid: 3e4c0102-a77a-47d8-a536-8c78178cea3e
service-name: nat
type: TC
loglevel: INFO
parent: 
rule:
 snat:

  entry:
   id  internal-net  external-ip
   0   10.0.0.0/24   1.2.3.5/32
 masquerade:
  enabled: false
 dnat:
 port-forwarding:

netgroup@giuseppe:~$ polycubectl nat1 rule del
netgroup@giuseppe:~$ polycubectl nat1 show
name: nat1
uuid: 3e4c0102-a77a-47d8-a536-8c78178cea3e
service-name: nat
type: TC
loglevel: INFO
parent: 
rule:
 snat:
 masquerade:
  enabled: false
 dnat:
 port-forwarding:
```

As I commented in the Nat.cpp file, I have some doubts about the subscribe but I have done both tests and various tests and everything seems to work.

### DIFF
--- a/src/services/pcn-nat/src/Nat.cpp
+++ b/src/services/pcn-nat/src/Nat.cpp
@@ -141,6 +141,7 @@ std::shared_ptr<Rule> Nat::getRule() {
 }
 
 void Nat::addRule(const RuleJsonObject &value) {
+  logger()->info("Adding rule");
   rule_ = std::make_shared<Rule>(*this, value);
 }
 
@@ -150,7 +151,11 @@ void Nat::replaceRule(const RuleJsonObject &conf) {
 }
 
 void Nat::delRule() {
-  rule_ = nullptr;
+  rule_ = nullptr; //In this way we can call the destroyers of the rules Snat, Dnat, Port-Forwarding
+  // We need an "empty" rule_ like the one created at initialization in the constructor. 
+  // Otherwise there is an error in the display of the nat after the deletion. As in issue #352
+  RuleJsonObject empty_rule;
+  rule_ = std::make_shared<Rule>(*this, empty_rule); 
 }
 
 std::shared_ptr<NattingTable> Nat::getNattingTable(


src/services/pcn-nat/src/Nat.cpp
```diff
@@ -150,7 +151,23 @@ void Nat::replaceRule(const RuleJsonObject &conf) {
...
+  // It seems to be the most logical choice, following the same logic used in the constructor of the Nat. 
+  // The NAT needs to know the IP address of the parent in order to perform the masquerate rules
+  ParameterEventCallback cb = [&](const std::string &parameter, const std::string &value) {
+    external_ip_ = utils::get_ip_from_string(value);
+    logger()->debug("parent IP has been updated to {}", external_ip_);
+    if (rule_->getMasquerade()->getEnabled()) {
+      rule_->getMasquerade()->inject(utils::ip_string_to_nbo_uint(external_ip_));
+    }
+  };
+  subscribe_parent_parameter("ip", cb);
```
FedeParola, line 170: Good job @pinoOgni.
I don't think this part is needed, the subscribe in the constructor should still be valid, can you test it?

FedeParola (changes requested):

src/services/pcn-nat/src/Nat.cpp
pinoOgni: Hi @FedeParola thanks for the reply. 

Yesterday I did some tests with and without the subscribe and it doesn't seem to change anything and everything works. I just wanted a second point of view, at this point I remove the subscribe.

FedeParola (approved):

Issue #350, opened 2020-10-23
[BUG] segmentation fault deleting a secondary ip that does not exist of a port of a router


pinoOgni:
## Describe the bug

Segmentation fault when trying to delete a secondaryip address in the format "ip-address/N" that does not exist.

## To Reproduce

In one terminal:
```
~  polycubectl router r1 
~  polycubectl r1 ports p1
~  polycubectl r1 ports p1
~  polycubectl r1 ports p1 set ip=10.0.0.1/24            
~  polycubectl r1 ports p1 secondaryip del 10.0.0.254/24 
Delete http://localhost:9000/polycube/v1/router/r1/ports/p1/secondaryip/10.0.0.254%2F24/: EOF
```
In another terminal:

```
~ sudo polycubed   
...
[2020-10-26 14:37:22.106] [Router] [r1] [info] Creating Router instance
[2020-10-26 14:37:33.637] [Router] [r1] [info] Added new port: p1 (index: 0) [mac: 56:21:c9:fb:d4:81 - ip: ]
[2020-10-26 14:38:33.520] [Router] [r1] [info] Added route [network: 10.0.0.1/32 - nexthop: 0.0.0.0 - interface: p1]
[2020-10-26 14:38:33.520] [Router] [r1] [info] Added route [network: 10.0.0.0/24 - nexthop: 0.0.0.0 - interface: p1]
[1]    15660 segmentation fault  sudo polycubed
```

## Expected behavior
Maybe a "warning" to the user instead of "segmentation fault". 

## Please tell us about your environment: 
1. OS details: cat /etc/-release 
```
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=20.04
DISTRIB_CODENAME=focal
DISTRIB_DESCRIPTION="Ubuntu 20.04.1 LTS"
NAME="Ubuntu"
VERSION="20.04.1 LTS (Focal Fossa)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 20.04.1 LTS"
VERSION_ID="20.04"
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
VERSION_CODENAME=focal
UBUNTU_CODENAME=focal

```

2. Kernel details: uname -a
```
[...]

PR #385, opened 2021-07-02
A possible solution to solve bug #350

pinoOgni:
This PR fixes bug #350  

The error was that before deleting the secondaryip, (by calling the `delete_router_ports_secondaryip_by_id` method which calls `delSecondaryip`), the method `read_router_ports_secondaryip_by_id` was called which inside calls `getSecondaryip(ip)`. Here  the case in which a secondaryip was not present was not handled (both because does not exist or because the port has no secondaryip or even the main ip address).

In the proposed solution, the `getSecondaryip` method returns nullptr if it does not find the secondaryip and then in the `read_router_ports_secondaryip_by_id method` an empty `PortsSecondaryipJsonObject` is returned; doing so will call the `delete_router_ports_secondaryip_by_id` (like a normal execution) method which will handle the error displaying it to the user.

Here there is an example:
```
netgroup@giuseppe:~$ polycubectl router r1
netgroup@giuseppe:~$ polycubectl r1 ports p1
netgroup@giuseppe:~$ polycubectl r1 ports p1 secondaryip del 1.1.1.3/24
Address does not exist (or it is not a secondary address)

netgroup@giuseppe:~$ polycubectl r1 ports p1 set ip=1.1.1.1/24
netgroup@giuseppe:~$ polycubectl r1 ports p1 secondaryip del 1.1.1.3/24
Address does not exist (or it is not a secondary address)

netgroup@giuseppe:~$ polycubectl r1 ports p1 secondaryip add 1.1.1.100/24
netgroup@giuseppe:~$ polycubectl r1 ports p1 secondaryip add 1.1.1.99/24
netgroup@giuseppe:~$ polycubectl r1 ports p1 secondaryip del  1.1.1.99/24
netgroup@giuseppe:~$polycubectl r1 ports p1 secondaryip del  1.1.1.99/24
Address does not exist (or it is not a secondary address)
```

### DIFF
--- a/src/services/pcn-router/src/Ports.cpp
+++ b/src/services/pcn-router/src/Ports.cpp
@@ -237,6 +237,10 @@ std::shared_ptr<PortsSecondaryip> Ports::getSecondaryip(const std::string &ip) {
     if (p.getIp() == ip)
       return std::make_shared<PortsSecondaryip>(p);
   }
+  // It can happen in 3 cases: the port has no IP, 
+  // the port has secondaryip but not the one sought, the port has 0 secondaryip
+  throw std::runtime_error(
+        "Address does not exist or it is not a secondary address or the port has no ip address");
 }
 
 std::vector<std::shared_ptr<PortsSecondaryip>> Ports::getSecondaryipList() {


FedeParola (changes requested):
Hi @pinoOgni,
unless strictly necessary I would avoid editing the `RouterApiImpl.cpp` that is auto-generated and should not be modified. It should be enough to just throw an exception in the `Ports::getSecondaryip()` method if the secondary ip is not found. Can you test it?

pinoOgni:
> Hi @pinoOgni,
> unless strictly necessary I would avoid editing the `RouterApiImpl.cpp` that is auto-generated and should not be modified. It should be enough to just throw an exception in the `Ports::getSecondaryip()` method if the secondary ip is not found. Can you test it?

Hi @FedeParola, you are right, I did not remember the constraint of not modifying the `RouterApiImpl.cpp` file. In any case I had tried to throw an exception in the `Ports::getSecondaryip` method but in doing so, the error message is not displayed to the user (as happens for example in the case of the `Ports::addSecondaryip`), but, there is only one error of this type (the throw error is not really seen):

```
netgroup@giuseppe:~$ polycubectl router r1
netgroup@giuseppe:~$ polycubectl r1 ports p1
netgroup@giuseppe:~$ polycubectl r1 ports p1 secondaryip del  1.1.1.99/24
{"ietf-restconf:errors":{"error":[{"error-type":"application","error-tag":"data-missing"}]}}
````

This solves the segmentation error since the `toJsonObject()` method is not called but I don't know if that's enough. What do you think about it? Do I push this last change or try to figure out how to make the user see an error? 
Although sadly, the only way I had found was to edit the `RouterApiImpl` file.

FedeParola:
I see, the behavior is strange...
When removing a port that does not exist for example we obtain the same json error:
`{"ietf-restconf:errors":{"error":[{"error-type":"application","error-tag":"data-missing"}]}}`
However when deleting a route that does not exist, we get the plain exception message:
`Cannot find a matching route.`
We should investigate why there are two different responses and possibly select just one.
For now however I would just fix the bug as we said, since it is compatible with the current behavior of Polycube.

pinoOgni:
> I see, the behavior is strange...
> When removing a port that does not exist for example we obtain the same json error:
> `{"ietf-restconf:errors":{"error":[{"error-type":"application","error-tag":"data-missing"}]}}`
> However when deleting a route that does not exist, we get the plain exception message:
> `Cannot find a matching route.`
> We should investigate why there are two different responses and possibly select just one.
> For now however I would just fix the bug as we said, since it is compatible with the current behavior of Polycube.

OK, I agree. I update the changes.

FedeParola (approved):

Issue #255, opened 2019-12-11
[Documentation] Document how to debug polycube


palexster:
## Is your feature request related to a problem? Please describe.
In order to simplify the debug of Polycube we should document how it can be run in debug mode.
Hint: 
* Enable the debug (https://stackoverflow.com/questions/7724569/debug-vs-release-in-cmake) and then use gdb gdb --args polycubed (don't forget to install after compiling)

frisso:
Done with #386 .

PR #386, opened 2021-07-09
doc to debug with VS Code and CLion (possible solution to #255)

pinoOgni:
This PR tries to provide a little guide to debug Polycube, using VS Code or CLion as IDE and tries to solve issue  #255 

Even though I've tried it several times, can anyone take a look? Thank you!
@frisso @FedeParola 

Comments on how to improve or modify it are welcome.

### DIFF
--- a/Documentation/developers/debugging.rst
+++ b/Documentation/developers/debugging.rst
@@ -126,11 +126,8 @@ Debugging using Integrated Development Environment (IDE)
 VS Code 
 ***************
 
-When compiling a project, by default the compiler creates all the produced files in the ``../build`` directory. 
-
-1. Polycube uses cmake to create the compilation environment (Makefiles, etc), which, by default is configured to create the executables in ``Release`` mode.
+Polycube uses ``cmake`` to create the compilation environment (Makefiles, etc), which, by default is configured to create the executables in ``Release`` mode.
 To debug your code, you must compile Polycube in ``Debug`` mode, with the following commands (follow also `here <https://stackoverflow.com/questions/7724569/debug-vs-release-in-cmake>`_):
-
 
   :: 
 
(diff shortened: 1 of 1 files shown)


Documentation/developers/debugging.rst
```diff
@@ -116,3 +117,97 @@ Usage example:
...
+
+
+Debugging using Integrated Development Environment (IDE)
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+
+VS Code 
+***************
+
+When compiling a project, by default the compiler creates all the produced files in the `../build` directory.
```
frisso, line 129: I would rephrase everything this way (but I'm not sure it's correct):

Polycube uses `cmake` to create the compilation environment (Makefiles, etc), which, by default is configured to create the executables in `Release` mode.
To debug your code, you must compile Polycube in `Debug` mode, with the following commands:

============
Additional comment: do we really need to set the breakpoints before compiling?

Documentation/developers/debugging.rst
```diff
@@ -116,3 +117,97 @@ Usage example:
...
+There are two equivalent ways to debug Polycube with CLion:
+
+1. Run CLion with sudo (Polycube needs root permissions)
+2. Set breakpoints and build/install Polycube (as explained above)
+3. At this point there are two possibilities:
+
+   a. Use the CLion debugger: at the top right select Debug 'polycube'; or
+   b. Run Polycube in a terminal like that `sudo polycubed --loglevel = DEBUG` in order to enable the Polycube Debug logs and then from CLion, go to Run->Attach to Process.. and search for "polycubed"
+
+4. At this point, from another terminal, just use polycubectl to interact with Polycube
```
frisso, line 212:
```suggestion
4. At this point, from another terminal, just use `polycubectl` to interact with Polycube.
```

Documentation/developers/debugging.rst
pinoOgni: Hi @frisso, yes, it is worth adding these two lines.  I'll update immediately.

Regarding the compilation before setting the breakpoints, I had some doubts about it, indeed now I have done other tests and it works even if they are set during debugging.

 Thank you!

Documentation/developers/debugging.rst
pinoOgni: Done! I also fixed two other things.

frisso (approved):

frisso:
Great work, Giuseppe!

PR #387, opened 2021-07-21
Documentation and comments to code to solve #42

### LINKED ISSUE #42: Document polycubed rest API
`polycubed` implements a rest API [1], these are some of the available functions: 
- Add, show and get services
- Get full list of cubes
- Get topology
- Get netdevs in the system
- Connect & disconnect cubes
- Get version

This API is not documented [2] 

[1] https://github.com/polycube-network/polycube/blob/master/src/polycubed/src/rest_server.h#L60
[2] https://github.com/polycube-network/polycube/blob/master/Documentation/polycubed/polycubed.rst#rest-api

pinoOgni:
This PR tries to provide documentation about Polycube's REST API and also to fix #42 

In this PR:

* a new file has been added that documents the REST API
* some comments have been added to the rest_server methods and other comments have been improved
* the logrequest() method has been fixed (It was not displaying the requested HTTP verb correctly)
* .gitignore file has been updated with the addition of the "Debug" directory


FedeParola (commented):
Hi @pinoOgni, overall good job.
I would however try to describe the rest API from a user point of view. For example inside the table in the documentation I would replace the `REST SERVER METHOD` column with an `REST ENDPOINT` column (with entries such as `/cubes/<cube_name>`) so that an user trying to reach the API knows what URL to use. The internal implementation of the API can be deduced from source files (that you commented) in my opinion. We could also consider having a user level description of the API in one file and some developer related information in another one. One additional part I would add is an high level description of how the API related to a service is generated. For example if a service has a list called `route` with key `ip` in the data model then there will be a `/route/<ip>` endpoint.
This is just an idea, let me know what you think.

pinoOgni:
> Hi @pinoOgni, overall good job.
> I would however try to describe the rest API from a user point of view. For example inside the table in the documentation I would replace the `REST SERVER METHOD` column with an `REST ENDPOINT` column (with entries such as `/cubes/<cube_name>`) so that an user trying to reach the API knows what URL to use. The internal implementation of the API can be deduced from source files (that you commented) in my opinion. We could also consider having a user level description of the API in one file and some developer related information in another one. One additional part I would add is an high level description of how the API related to a service is generated. For example if a service has a list called `route` with key `ip` in the data model then there will be a `/route/<ip>` endpoint.
> This is just an idea, let me know what you think.

Hi @FedeParola, thanks for the review, I uploaded the wrong table (I had some problem with the rst file), the right one has the endpoint used as the first column as you rightly said, I'll update immediately.

FedeParola (approved):

Documentation/developers/rest-api.rst
```diff
@@ -0,0 +1,105 @@
+Rest API
+=====================================
+
+The Polycube rest API is implemented in the [rest_server.cpp](https://github.com/polycube-network/polycube/blob/master/src/polycubed/src/rest_server.cpp),
```
frisso, line 4:
```suggestion
Polycube REST API is implemented in [rest_server.cpp](https://github.com/polycube-network/polycube/blob/master/src/polycubed/src/rest_server.cpp), 
```

frisso (approved):

PR #388, opened 2021-07-23
Create codeql-analysis.yml

frisso:


frisso:
Let's do code analysis seriously, when we really have some spare time to dedicate to this task.

Issue #234, opened 2019-11-13
[BUG] Remove references of pcn_pkt_log()


frisso:
## Describe the bug
A previous commit left some references of the pcn_pkt_log() function, which is no longer used (use span more instead).

## Additional context
Current references can be found by looking at:
https://github.com/polycube-network/polycube/search?q=pcn_pkt_log&unscoped_q=pcn_pkt_log

frisso:
Done in #389

PR #389, opened 2021-07-27
Solved issue #234

pinoOgni:
This PR completes the work of removing pcn_pkt_log() references from code and documentation, resolving #234


frisso:
@pinoOgni I would suggest to add a small description how you can capture a packet with tcpdump, through the mechanism of the "shadow ports". This looks very nice to me, but it's poorly documented. The "debugging.rst" file looks perfect to put a small text, with an example, showing how to capture packets and see what's going on.

frisso (approved):

PR #399, opened 2021-08-20
Update the polycubed starting string

frisso:
Polycube doesn't work if started with "&", at least the first time this is done after booting up.
Much better to use the "-d" flag.


pinoOgni (approved):
lgtm

PR #407, opened 2022-01-20
fixed Lbrp::delService handling of service backends deletion

ekoops:
Hi folks! If we try to delete a service without backends, calling directly Service::removeServiceFromKernelMap will generate an exception, so is needed to call Service::delBackendList (that performs a check on the number of backends and performs the cleanup in the right way) in order to correctly delete a service.

### DIFF
--- a/src/services/pcn-loadbalancer-rp/src/Lbrp.cpp
+++ b/src/services/pcn-loadbalancer-rp/src/Lbrp.cpp
@@ -389,7 +389,7 @@ void Lbrp::delService(const std::string &vip, const uint16_t &vport,
 
   auto service = getService(vip, vport, proto);
 
-  service->removeServiceFromKernelMap();
+  service->delBackendList();
   service_map_.erase(key);
 }
 
--- a/src/services/pcn-loadbalancer-rp/src/Service.cpp
+++ b/src/services/pcn-loadbalancer-rp/src/Service.cpp
@@ -536,7 +536,7 @@ void Service::delBackend(const std::string &ip) {
 }
 
 void Service::delBackendList() {
-  if (service_backends_.size() != 0) {
+  if (!service_backends_.empty()) {
     service_backends_.clear();
     backend_matrix_.clear();
     removeServiceFromKernelMap();


src/services/pcn-loadbalancer-rp/src/Lbrp.cpp
```diff
@@ -389,7 +389,7 @@ void Lbrp::delService(const std::string &vip, const uint16_t &vport,
 
   auto service = getService(vip, vport, proto);
 
-  service->removeServiceFromKernelMap();
+  service->delBackendList();
```
frisso, line 392: @ekoops What is this method doing? Aren't we deleting too much here? (we should delete only a given service, while here we're deleting the entire backend list). Thanks!

ekoops: The code of the method Service::delBackendList is the following:
```
if (!service_backends_.empty()) {
    service_backends_.clear();
    backend_matrix_.clear();
    removeServiceFromKernelMap();
}
```
service_backends and backend_matrix are two user space maps associated to the specific service:
```
std::unordered_map<std::string, ServiceBackend> service_backends_;
std::map<std::string, std::vector> backend_matrix_;
```
 and have to be cleaned up only in the case the service have some associated backend. The delBackendList method also call the method removeServiceFromKernelMap which is in charge of removing from the kernel space eBPF map all the entry related to the specific service. Indeed, if we look at the implementation of ServiceServiceFromKernelMap, the code is the following:
```
void Service::removeServiceFromKernelMap() {
  logger()->trace("Removing all elements from service map");

  auto services_table = parent_.get_hash_table<vip, backend>(EBPF_SERVICE_MAP);

  for (uint16_t index = 0; index < backend_size_ * BACKEND_REPLICAS; index++) {
    vip service_key{
        .ip = utils::ip_string_to_nbo_uint(getVip()),
        .port = htons(getVport()),
        .proto = htons(Service::convertProtoToNumber(getProto())),
        .index = index,
    };

    services_table.remove(service_key);
  }
}
```
We can clearly see that this method try to delete the backends associated to the service (notice getVip, getVport and getProto calls) at each index. The problem is that if this is method is called and the given service key is not present, service_table.remove(service_key) will raise an exception, and this happens in case of a raw call of removeServiceFromKernelMap for a service with no backends. A call to delBackendList will ensure that the two user space map are correctly cleaned up, and the check on the presence of backends will ensure that no exception is raised during the deletion of backends associated to the service in the kernel map.

frisso (approved):

